Issue #20, opened 2017-02-11
Add mysql service


pirog:

serundeputy:
@pirog this is a first pass at adding the `mysql` service: https://github.com/kalabox/lando/pull/25

Here is the output:

```
geoff@yep mysql (20/add-mysql-service) $ lando start
warn: mysql is not a supported service.
Starting mysql_data_1
Starting mysql_unisondata_1
Starting mysql_html_1
Starting lando_proxy_1

BOOMSHAKALAKA!!!

Your app has started up correctly.
Here are some vitals:

 NAME        mysql
 LOCATION    /Users/geoff/code/lando/examples/mysql
 CONTAINERS  data, html, unisondata
 URLS        https://localhost:32793
             http://localhost:32794
             http://mysql.kbox.site
             https://mysql.kbox.site
```

Maybe you can help me iterate and find the missing pieces.

thanks,
~Geoff

serundeputy:
Updated PR https://github.com/kalabox/lando/pull/25 to register the `mysql` service.
Rebased to one commit.

PR #25, opened 2017-02-11
Issue #20: Adding mysql service.

serundeputy:
Needs work.


pirog:
@serundeputy def a good start here. i think the big thing that is missing is "registering" the service. check out: https://github.com/kalabox/lando/blob/master/plugins/lando-services/lib/bootstrap.js#L31

I'll be trying to finish up the project scaffolding this weekend so we can be set up for MAX DEV in the coming months!

PR #124, opened 2017-06-26
fixing the git command - to clone drupal7 from Pantheon

shaal:

- [x] My code includes a test if applicable ([see here](https://docs.lndo.io/dev/testing.html))

improve our process.


docs/recipes/pantheon.md
```diff
@@ -45,7 +45,7 @@ git clone https://github.com/kalabox/lando.git
 cd lando/examples/pantheon
 
 # Clone drops-7 into a nested docroot
-git clone git@github.com:pantheon-systems/drops-7.git web
+git clone https://github.com/pantheon-systems/drops-7.git web
```
uberhacker: My guess is the reason this change is necessary is because your Pantheon account doesn't have the ssh pubkey yet?  I believe Terminus has commands now that let you manage ssh keys.  Maybe add a key and it will work?

shaal: I thought it will be better to have the simplest recipe possible for beginners, with fewer requirements.
This way - users don't need to get a pantheon ssh pubkey before they can see the recipe working.

uberhacker: Users don't need to add their own ssh key if Terminus does it as part of the process.

pirog: i'm guessing the docs around this will change soon but ill merge for now

Issue #149, opened 2017-07-06
Write a tutorial on how to run behat with lando (and in CI with lando)


pirog:

serundeputy:
I started on this today.  Here is what I've got:

* started with Pantheon `example-drops-8-composer` as it has a composer with Behat and friends in it.

```bash
composer create-project pantheon-systems/example-drops-8-composer behat-d8
lando init behat-d8
lando drush si --db-url=mysql://drupal8:drupal8@database/drupal8
```

In `tests/behat-pantheon.yml` I had to add to the `drush` section beneath the `api_driver` section like so:

```yaml
#
# behat.yml file for testing on Pantheon.
#
default:
  suites:
    default:
      paths:
        - %paths.base%/features
        - %paths.base%/site-features
      contexts:
        - FeatureContext
        - Drupal\DrupalExtension\Context\DrupalContext
        - Drupal\DrupalExtension\Context\MinkContext
  extensions:
    Behat\MinkExtension:
      goutte: ~
      selenium2: ~
      files_path: './data-files'
    Drupal\DrupalExtension:
      blackbox: ~
      api_driver: 'drush'
      drush:
        alias: 'local'
```

Behat tries to write to `/tmp` files, but in the container they are owned by `root` so I changed the ownership there to `www-data`:

```bash
lando ssh
chmod -R www-data:www-data /tmp
```

We'll want a way to have that set up already for the user `extras` key in `.lando.yml`?

serundeputy:
The `pantheon-systems/example-drops-8-composer` repo does not include a `.htaccess` file so I had to copy one in to make clean URLs work and the site be useable.

serundeputy:
PR: https://github.com/kalabox/lando/pull/253

Issue #147, opened 2017-07-06
Write a tutorial on how to run lando in a CI envrionment


pirog:
Let's write a tutorial (https://github.com/kalabox/lando/tree/master/docs/tutorials) that a user can follow so that when they are done reading they have a good idea of how to set up lando in travis or circle ci so they can super streamline their devops. presumably this would include

1. installing kalabox on travis/circle
2. running a lando start comand
3. having lando handle dependency installation 
4. having lando handle task running

(a basic code linting and code standards check here is probably best)

serundeputy:
PR: https://github.com/kalabox/lando/pull/246

serundeputy:
@pirog updated with `examples/lando-and-ci/.lando.yml` full example file and included in the bottom of the tutorial.

PR #253, opened 2017-08-02
#149 lando ci and behat

serundeputy:


pirog:
@serundeputy this is all in #147 now

Issue #292, opened 2017-08-21
[DX] Add a db-export command.


serundeputy:
As a developer sometimes I want to get a database out of a lando app.  It can be done via `lando ssh` or SequelPro if you know how to `msyql` on command line or connect SequelPro to lando and that is fine and good, but the `db-export` command makes it a little easier / gives another way.

I've added:

```bash
lando db-export
```

to directly export the db to the `/app` directory; modeled after the `db-import` lando pattern.

Here is the info from the commands list:

```bash
geoff@yep www (master) $ lando.dev
Usage: /usr/local/bin/lando.dev <command> [args] [options] [-- global options]

Commands:
  composer                 Run composer commands
  config                   Display the lando configuration
  db-export                Export a datbase. Resulting {DB_NAME}.TIMESTAMP.gz file is in /app.
  db-import <file>         Import <file> into database. File is relative to approot.
```

serundeputy:
PR: https://github.com/kalabox/lando/pull/293
I'll work on some docs.

serundeputy:
PR and docs updated

PR #293, opened 2017-08-21
#292: Add db-export command for mysql.

serundeputy:


pirog:
@serundeputy we've got options surfaced as part of the `lando db-export` command but they are not handled when passed to the `db-export` script. Probably worth grabbing the option handling code from the `db-import` script and reusing that here!

serundeputy:
@pirog added the option handling.

pirog:
@serundeputy nice! ill try to give this the ole look-see early next week

PR #299, opened 2017-08-24
Add tooling + events example to the docs

jeffm2001:
Based on discussion last week in slack, this adds an example to the docs for using events to create tooling that runs on multiple services. I added it to the Tooling tutorial because I don't think I would have found it if it was under events.


pirog:
@jeffm2001 nice! thanks for doing this. Will review and merge soon!

pirog:
@jeffm2001 thanks again!

PR #365, opened 2017-09-27
Use service names for Behat config.

dustinleblanc:


pirog:
@dustinleblanc thanks!

PR #370, opened 2017-09-28
proxy docs: add info on connecting from other devices on local network

ptmkenny:
- [X] I have pinged @pirog/@reynoldsalec/@serundeputy when I think my code is ready for prime time.

@pirog @serundeputy Here's an addition to the docs to address connecting to lando via devices on a local network; this was requested in the kalabox issue queue: https://github.com/kalabox/kalabox/issues/1807

I just set this up for myself today so I thought I would share my notes.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -36,6 +36,7 @@
 *   [Setting up Additional Events](config/events.md)
 *   [Using NodeJS Frontend Tooling](tutorials/frontend.md)
 *   [Accessing Your Services Externally](tutorials/external-access.md)
+*   [Accessing Lando from Other Devices](tutorials/access-by-other-devices.md)
 *   [Importing Databases](tutorials/db-import.md)
 *   [Exporting Databases](tutorials/db-export.md)
 *   [Using Composer to Manage a Project](tutorials/composer-tutorial.md)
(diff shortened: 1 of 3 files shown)


pirog:
@ptmkenny this is great but it might be worth mentioning and linking to the `lando share` docs since that is also a pretty good way to do this. Thoughts?

ptmkenny:
@pirog Darn, I overlooked the `lando share` command, that's awesome.  I added it to the top of the documentation and put it as the recommended method since it works over the Internet and not just a local network, and I put the local IP method second if people don't want the app exposed to the Internet.

pirog:
@ptmkenny yeah `lando share` is one of those hidden gems :)

So i think this looks great. My only question is do you think it makes more sense to have this be part of the proxy documentation or its own separate tutorial?

ptmkenny:
@pirog  It might be easier to find if this was its own section with a title like "Access from Other Devices".

pirog:
@ptmkenny exactly. what i would do is spin up a new file called `access-other-devices.md` (or something similar) and drop it into the tutorial folder in the docs. Then i would update the `SUMMARY.md` to include this as a tutorial.

ptmkenny:
@pirog Done!

pirog:
@ptmkenny awesome! thanks so much! really appreciate all you've done to help the project!

Issue #376, opened 2017-10-02
`lando` with no params returns error should return possible commands.


serundeputy:
Since spinning up `v3.0.0-beta.18` using the command `lando` with no parameters returns an error:

```bash
geoff@yep lightupthemap (11-rest-create-lights) $ lando.dev
/Users/geoff/code/lando.dev/node_modules/yargs-parser/index.js:748
  var result = parse(args.slice(), opts)
                          ^

TypeError: args.slice is not a function
    at Parser (/Users/geoff/code/lando.dev/node_modules/yargs-parser/index.js:748:27)
    at Object.exports.parseGlobals (/Users/geoff/code/lando.dev/lib/tasks.js:75:10)
    at Object.<anonymous> (/Users/geoff/code/lando.dev/lib/tasks.js:93:22)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
```

pirog:
@serundeputy nice find here. rolling a hotfix release now

PR #377, opened 2017-10-02
#376: Allow lando command to work with no parameters; return possible…

serundeputy:


lib/tasks.js
```diff
@@ -70,12 +70,10 @@ exports.parseGlobals = function() {
...
   }
-
-  // Return the globalz
-  return parse(globz, {
-    alias: {'verbose': ['v'], 'help': ['h']},
-    count: ['v']
-  });
+  else {
+    globz = [];
+    return globz;
```
pirog: @serundeputy i dont think this line is needed?

serundeputy: w/out the `return` line it doesn't work but with it it breaks at least `lando version`

pirog: @serundeputy right, sorry, so i think you want to add the `else` to the filtering block not the return block

Issue #381, opened 2017-10-04
Set Drupal Console on by default for D8 recipes?


mmenavas:
Has anyone been able to run Drupal Console commands? 

Per the [documentation](https://docs.devwithlando.io/tutorials/drupal8.html#tooling) for Drupal 8, I should be able to run `lando drupal`; however, after running the commands below, I get `Unknown argument: drupal`.

```bash
git clone -b 8.4.x https://github.com/drupal/drupal.git d8
cd d8
lando init --recipe drupal8
lando start
lando composer install
lando drupal
```

I'm currently running Lando v3.0.0-beta.19 on macOS Sierra (10.12.6).  

My auto-generated yaml file looks like this
```yml
name: d-8
recipe: drupal8
config:
  webroot: .
```

The output of the same command with verbose mode on:

```bash
lando drupal -- -vvvv
verbose: Loading event pre-bootstrap priority 1
info: Bootstrap starting...
[...]

PR #383, opened 2017-10-04
#381: Automate install of site local drupal console in drupal8 recipe.

serundeputy:


serundeputy:
@pirog not sure if this is the best way to do this, but It appears that `drupal` console now requires a site local install to work.  So I've added them composer command to require it and run `composer install` if a `composer.json` file is present.

pirog:
@serundeputy i'm a little conflicted about this. I definitely think we want to auto install the global dependency part of console (like we do for drush or gulp), but it might be tricky to assume what the user wants for their local console. Which is to say it would be EASIEST to just tell users they need to install console via their apps `composer.json` as indicated in the console install docs. Maybe i'm just not understanding this.

serundeputy:
closing so we don't blow up peoples site `composer.json` file.  We can update the docs to provide some context and clarification.

PR #390, opened 2017-10-07
#381: Include drupal console by default and install it site-local if …

serundeputy:
…not already required.


pirog:
@serundeputy and i decided to provide deeper "auto composer install" functionality for all php-y things which as consequence would cover this PR as well

PR #397, opened 2017-10-10
adds troubleshooting for Linux

alexfornuto:
Documents some issues that through me for a loop while installing Lando from source.

- [ ] I have pinged @pirog when I think my code is ready for prime time.

### DIFF
--- a/docs/changelog/2017.md
+++ b/docs/changelog/2017.md
@@ -20,7 +20,7 @@ v3.0.0-beta.18 - [October 1, 2017](https://github.com/lando/lando/releases/tag/v
 v3.0.0-beta.17 - [September 9, 2017](https://github.com/lando/lando/releases/tag/v3.0.0-beta.17)
 ------------------------------------
 
-**SERIOUSLY, READ THE DOCS: https://docs.devwithlando.io/**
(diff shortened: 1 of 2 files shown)


alexfornuto:
@pirog Does this count as a ping?

pirog:
@alexfornuto this looks great. only suggestion is instead of having a troubleshooting section under the "install from source" section, you bump it up to the linux section as some sort of "caveats" subsection.

what do you think?

alexfornuto:
@pirog I've move it as suggested, but I'm not sure if The header is still at the right level. This document uses two different forms of Header markdown (underlines `---` `===` and octothorpes `###`) which makes it hard to determine nesting level.

pirog:
As far as i know 

```
= -> h1
- -> h2
### -> h3
#### - > h4
```

so im guessing you want `###`?

alexfornuto:
@pirog You could standardize it as:

```
# => H1
## => H2
### => H3
#### => H4
```

But that may just be a personal preference.

pirog:
Yeah, i know that is a common way to do it. 

Personally i find using `==` and `--` for the `h1` and `h2` to be a bit more readable when you are looking at lots of markdown files all the time... as a sadly am :(

PR #416, opened 2017-10-17
info on fixing file uploads in Drupal

ptmkenny:
- [X] I have pinged @pirog/@reynoldsalec/@serundeputy when I think my code is ready for prime time.


pirog:
@ptmkenny this is super awesome but i'm wondering if it might be better to have these these live in the "known issues" section of our docs under a title like *Drupal file upload issues on Windows*.

What do you think? Are you open to adding a separate `.md` file into `docs/issues` folder and then adding it into the `known issues` section of `SUMMARY.md`?

pirog:
@ptmkenny i took this the rest of the way for you!

PR #452, opened 2017-10-31
Update usage.md to include Pull command

ccharlton:
Added a quick line for the `pull` command.


pirog:
@ccharlton im curious to hear more about your thinking behind this. 

As of right now `lando pull` is a `pantheon` recipe specific command so i'm not sure it belongs in the general CLI usage docs

PR #461, opened 2017-11-02
[docs] update indenting in `build` section

nullvariable:
build command appears to be indented incorrectly, I've updated the relevant lines

- [ n/a] My code passes relevant CI status checks
- [ n/a] My code includes a test if applicable ([see here](https://docs.devwithlando.io/dev/testing.html))
- [ n/a] My code includes an update to the `CHANGELOG` ([see here](https://github.com/kalabox/lando/tree/master/docs/changelog))
- [ x] My code includes documentation updates if relevant.
- [ n/a] I have pinged @pirog/@reynoldsalec/@serundeputy when I think my code is ready for prime time.

improve our process.

### DIFF
--- a/docs/tutorials/frontend.md
+++ b/docs/tutorials/frontend.md
@@ -20,8 +20,8 @@ recipe: lamp
 services:
   node:
     type: node:6.10
-  build:
-    - "cd $LANDO_MOUNT && npm install"
+    build:
+      - "cd $LANDO_MOUNT && npm install"
 ```
 
 Note the `build` section. This bash command will automatically run when we start our app, installing any Node packages specified in our package.json file.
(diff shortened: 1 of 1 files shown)


pirog:
@nullvariable, i *think* the rest of the docs has two spaces for indents in `yaml` files so ... not exactly sure this is the best change? Can you explain more of your thinking here?

nullvariable:
@pirog Hey Mike, I probably could've explained a bit better :) If you copy/paste that code as is, it won't work because the `build:` step in the yaml file needs to be indented underneath a service. At least that was my experience when using it. It could be that I was doing it wrong in a different way.

Said another way, in the original example, `build:` is inline with `node:` causing it to think that `build:` is a new container instead of a build instruction for the node container. My commit just indents the build command two spaces so it's a child of the `node:` container.

pirog:
@nullvariable got it! i totally read this wrong. my bad and thanks!

PR #465, opened 2017-11-03
Update drupal8.md

AmyJuneH:
git clone -b 8.x https://github.com/drupal/drupal.git mysite

should be 

git clone -b 8.5.x https://github.com/drupal/drupal.git mysite

- [ ] I have pinged @pirog/@reynoldsalec/@serundeputy when I think my code is ready for prime time.

improve our process.

### DIFF
--- a/docs/tutorials/drupal8.md
+++ b/docs/tutorials/drupal8.md
@@ -15,7 +15,7 @@ Before you can use all the awesome Lando magic you need a Drupal 8 codebase with
 ```bash
 # Clone or extract your Drupal 8 site
 # See: https://github.com/drupal/drupal/tree/8.x
-git clone -b 8.x https://github.com/drupal/drupal.git mysite
+git clone -b 8.5.x https://github.com/drupal/drupal.git mysite
 
 # Or install with composer
 # When you lando init below you need to use `web` as your webroot


pirog:
@AmyJuneH @sk33lz oh interesting. i didn't know D8 was branching this way for minor versions, makes sense i guess. 

Sort of annoying that we will be forced to update this as we go... maybe there is a more robust solution EG instead of git cloning maybe we can curl/wget some sort of "latest" D8 package. 

I'll merge this in for now but if either of you have energy to change the docs to address the brittleness of this it would be much appreciated!

pirog:
Enjoy the beautiful American Paradise that is Worcester, MA!

jeffymahoney:
Nice :)
Here's what we can change the docs to (sorry, no time today to get the repo and create a proper PR right now.)

```
# See: https://github.com/drupal/drupal
git clone https://github.com/drupal/drupal.git mysite
```

Just leave off the -b option and after the clone, you'll be on 8.5.x. Also note that the link in the comment needs to be updated. The old link, https://github.com/drupal/drupal/tree/8.x , is outdated.

Peash :)

jeffymahoney:
And thanks for making this change. It really helps out people who are still learning the ropes.

PR #467, opened 2017-11-04
Fix indentation with services:node:build

MatthieuScarset:
There is an indentation typo. The `build` method needs to be indented under `services:node` instead of being at the same level as `node`.


pirog:
ick! super good catch here @MatthieuScarset. thanks a bunch!

Issue #464, opened 2017-11-03
Feature Request: Please add configuration to turn off the pantheon composer build upon 'lando pull'.


ozheller:
As a user using Lando and Pantheon, we primarily use the build artifacts that Pantheon provides. The build artifact contains composer.json and composer.lock files. When I do 'lando pull', during the pull process composer will rebuild the vendor folders. I don't want to automatically overwrite the vendor folder. I want the vendor folder on my local to match the pantheon repo.

The current easy solution is to delete the composer files before doing a lando pull. Although deleting the composer.jason and composer.lock files from the repo may have adverse effects when trying to update from pantheon's upstream.

Please add configuration so we may disable the composer portion of lando pull. 

[improvement]

serundeputy:
PR: https://github.com/lando/lando/pull/468

-----

@ozheller I could not get `composer install` to run on `lando pull`, but it does run on `lando start`, `lando restart`, and `lando rebuild.

The PR adds a `composerSwitch` key that a user can set to false to disable `composer install` on start operations.

```yml
# By default the operations `lando start`, `lando restart`, and
# `lando rebuild` run `composer install` on the app.
#
# The `composerSwitch` key allows users to disable this behavior and run
# `composer install` when and if you choose by setting a value of false.
#
# composerSwitch: false
```

PR #468, opened 2017-11-04
#464 Adding composerSwitch key to disable composer install on start ops.

serundeputy:


pirog:
@serundeputy merged this into the `beta23` branch so closing here

PR #469, opened 2017-11-06
Fix Drupal 6 Accessing services link

seantwalsh3:
The Accessing services (eg your database) from the host is currently pointing to http://docs.devwithlando.io/tutorials/frontend.html instead of https://docs.devwithlando.io/tutorials/access-by-other-devices.html


pirog:
@seantwalsh3 good catch here but i think we want to be linking to 
https://docs.devwithlando.io/tutorials/external-access.html instead of https://docs.devwithlando.io/tutorials/access-by-other-devices.html ?

seantwalsh3:
Sorry about that. I've starting using Lando as my primary local dev so will be submitting some other docs issues in the future.

pirog:
Most appreciated! They are def still a bit rough around the edges!

Issue #453, opened 2017-11-01
Nested webroot not recognised if set in pantheon.upstream.yml, but not pantheon.yml


dalin-:
Bug Report
==========

**Tell us about your setup**

What is your lando version and operating system?
v3.0.0-beta.22
OSX 10.12

**Tell us about your `.lando.yml`**

```yml
name: eam-d8
recipe: pantheon
config:
  framework: drupal8
  env: dev
  site: eam-d8
  id: bef78b81-d3af-4eff-b896-a93c46cd8e23
```

**Tell us about the command you were running**

```bash
lando init pantheon
lando start
# Then try to access site in the browser
```

**Tell us about the error you got**

Generic nginx 500 error

**Tell us generally about your bug**

But if I go to 
https://localhost:32793/web/core/CHANGELOG.txt
I see the text file

But if I 

```sh
cp pantheon.upstream.yml pantheon.yml
lando restart
```

Then everything works fine. 

**Tell us more**

Could better documentation have prevented this issue? if so what should we change?

No, it just sounds like if `pantheon.yml` is present, then `pantheon.upstream.yml` is ignored.  But instead it should override and add to the upstream values.

Does this bug prevent you from using lando?
No, I have this workaround.

serundeputy:
PR: https://github.com/lando/lando/pull/482

PR #482, opened 2017-11-09
#453: Respect pantheon.upstream.yml if pantheon.yml has nothing bette…

serundeputy:


pirog:
@serundeputy ditto! this is in the beta23 branch now! closing here and will push as part of that!

PR #485, opened 2017-11-10
Updating drupal8.md 

AmyJuneH:
Just leave off the -b option and after the clone, you'll be on 8.5.x. Also note that the link in the comment needs to be updated. The old link, https://github.com/drupal/drupal/tree/8.x , is outdated.


pirog:
@AmyJuneH provided more generic docs around spinning up a lando.yml but we might want to add this elsewhere!

Issue #520, opened 2017-11-20
Always getting 404s or empty 200s on Linux


anavarre:
**Tell us about your setup**

```
$ uname -ar
Linux thinkpad 4.13.0-16-generic #19-Ubuntu SMP Wed Oct 11 18:35:14 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

**What is your lando version and operating system?**

```
$ lando version
v3.0.0-beta.25
$ cat /etc/issue
Ubuntu 17.10 \n \l
```

**Tell us about your `.lando.yml`**

```
$ cat .lando.yml 
name: my-lando-app
recipe: drupal8
config:
  webroot: .
```

This is the default that comes when picking Drupal 8 in the init phase. The webroot corresponds to `~/Lando/` and it's writable by the user.

**Tell us about the command you were running**

`$ lando start`

**Tell us about the error you got**

None of the URLs provided (either localhost:port or domain name) work. I'm getting a 404 or 200 with the following:

```
Index of /

======================
Apache/2.4.10 (Debian) Server at localhost Port 32783
```

**Tell us generally about your bug**

I'm just trying to get a minimal D8 site working without doing anything else but using the defaults.

**Tell us more**

I'm trying to understand if there's something I should be doing differently. Reading the documentation didn't make obvious what steps I missed, if any. Here's a full run:

```
$ lando init
? What recipe do you want to use? drupal8
? Where is your webroot relative to the init destination? .
? What do you want to call this app? my-lando-app

NOW WE'RE COOKING WITH FIRE!!!

Your app has been initialized.
Now try running `lando start` to get rolling.

Here are some vitals:

 NAME    my-lando-app                                        
 RECIPE  drupal8                                             
 DOCS    https://docs.devwithlando.io/tutorials/drupal8.html 

$ lando start
landoproxyhyperion5000gandalfedition_proxy_1 is up-to-date
Recreating mylandoapp_appserver_1 ... 
Recreating mylandoapp_database_1 ... 
Recreating mylandoapp_appserver_1
Recreating mylandoapp_appserver_1 ... done
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 5988k  100 5988k    0     0  1582k      0  0:00:03  0:00:03 --:--:-- 1582k
[...]

PR #523, opened 2017-11-20
Clarify Lando docs

anavarre:
This should help people not fall into the trap of #520.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,8 @@ lando init
 lando start
 ```
 
+Note: Lando will not spin up a new codebase for you unless you init with either the `pantheon` or `github` method (see below). Otherwise, remember to always have an already existing project to init from.
+
 #### 3. Or pull from GitHub or Pantheon
 
 ```bash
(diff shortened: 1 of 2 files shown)


pirog:
@anavarre i think having this clarification is good but we might want to clarify the clarification. Lando WILL install a codebase for you if you use either the `pantheon` or `github` initialization methods. But a straight up `lando init` is intended for usage in an already existing codebase, i think this is the case we want to clarify. 

Let me know if you can make those changes and then i'll merge! Thanks for your help!

anavarre:
@pirog sure, I've clarified this. Does it make more sense as is?

pirog:
@anavarre yes! thanks so much for helping.

ujeshurun:
Sorry, I am a newcomer to this, so forgive me if I ask any silly questions. But just to get this understood, if I wanted to have lando spin up a new wordpress instance, I CANNOT SIMPLY RUN `lando init`?

serundeputy:
@ujeshurun that is correct.

You need to either download a new wordpress code base or a wordpress site you already have; `cd into the wordpress directory` then `lando init`

Issue #501, opened 2017-11-15
Add composer install build step to php recipes.


serundeputy:
Currently the Pantheon recipe runs `composer install` as part of the build process.  Let's move it from that recipe upstream to all PHPy recipes that inherit from `lamp/lemp`.

This will solve #381 and users can now gain there own control over running composer automatically or not by setting the ~~`composerSwitch`~~ `disableAutoComposerInstall` key to `true`.

pirog:
@serundeputy as a FYI i changed the name of the key 
https://github.com/lando/lando/blob/master/examples/pantheon/.lando.yml#L39

serundeputy:
Yep; I saw that thanks! #namingthingsishard;

jbertoen:
As an addition, wouldn't it be sweet if we could include some "global" libraries from the configuration of the .lando.yml to the build? Which would make sure that tools like drush or console that someone might want globally is available on the service.
We often are "tired of waiting" for composer install to be finished. 
http://jackmcdade.com/blog/tired-of-waiting
They for instance recommend to globally require prestissimo to speed up the composer install process. Works like a charm for our drupal setup btw.
something like 
`composerGlobalRequire:
- hirak/prestissimo
- drush/drush
- drupal/console`

pirog:
@jbertoen you can already do this! https://docs.devwithlando.io/services/php.html#lamp-example check out the composer key.

also worth mentioning that composer global usage is discouraged

PR #531, opened 2017-11-22
#501: Move composer auto install logic to lamp.js and let Drupals and…

serundeputy:
… Backdrop inherit it.

### DIFF
--- a/plugins/lando-recipes/lamp/lamp.js
+++ b/plugins/lando-recipes/lamp/lamp.js
@@ -337,25 +337,6 @@ module.exports = function(lando) {
     build.services = services(config);
     build.tooling = tooling(config);
 
-    // Determine the service to run cli things on
-    var unsupportedCli = (config.php === '5.3' || config.php === 5.3);
-    var cliService = (unsupportedCli) ? 'appserver_cli' : 'appserver';
-
-    // Build an additional cli container if we are running unsupported
-    if (unsupportedCli) {
-
-      // Build out a CLI container and modify as appropriate
-      var cliImage = 'devwithlando/pantheon-appserver:5.5-fpm';
-      build.services[cliService] = _.cloneDeep(build.services.appserver);
-      build.services[cliService].type = 'php:5.5';
-      build.services[cliService].via = 'cli';
-      build.services[cliService].overrides.services.image = cliImage;
-
-      // Remove stuff from appserver
-      delete build.services.appserver.build;
-
-    }
-
     // Check if the user specified the compserSwitch key to false
     var disableComposer = _.get(config, 'disableAutoComposerInstall', false);
     var composerJson = path.join(config._root, 'composer.json');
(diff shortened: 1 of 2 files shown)


plugins/lando-recipes/lamp/lamp.js
```diff
@@ -336,6 +337,36 @@ module.exports = function(lando) {
     build.services = services(config);
     build.tooling = tooling(config);
 
+    // Determine the service to run cli things on
+    var unsupportedCli = (config.php === '5.3' || config.php === 5.3);
```
pirog: @serundeputy this stuff is pretty pantheon specific. whats the thinking around moving this over to lamp??

plugins/lando-recipes/lamp/lamp.js
```diff
@@ -336,6 +337,36 @@ module.exports = function(lando) {
...
+      build.services[cliService].via = 'cli';
+      build.services[cliService].overrides.services.image = cliImage;
+
+      // Remove stuff from appserver
+      delete build.services.appserver.build;
+
+    }
+
+    // Check if the user specified the compserSwitch key to false
+    var disableComposer = _.get(config, 'disableAutoComposerInstall', false);
```
pirog: @serundeputy so i think we want this stuff to exist directly in the php service. the only thing the lamp recipe should be doing is checking for the `disableAutoComposerInstall ` setting in its `config` and then passing that down to the php service just like we do for the php version.

plugins/lando-recipes/lamp/lamp.js
serundeputy: @pirog if a user is using lemp and php 5.3 it will spin up a php appserver_cli container.  WE don't have to use the pantheon image for it; do you have a recommendation for a better image here?

pirog: @serundeputy composer works in php 5.3 so i assumed we'd just run it in the appserver without needed a cli container. We only use the cli container on pantheon in the 5.3 use case because terminus requires 5.5+

serundeputy: gotchya; I'm moving this back to the Pantheon recipe.

plugins/lando-recipes/lamp/lamp.js
serundeputy: I'm not sure how to do this one ... the php service `plugins/lando-services/php/php.js` doesn't have a 

```js
var build = function(...
```
step.

Should I add one?

pirog: im not sure it needs `var build`, i think you probably just want to do what our composer key does
https://github.com/lando/lando/blob/master/plugins/lando-services/php/php.js#L332

pirog: @serundeputy so i think it should go in the `services` function and work similarly to how we handle the `composer` global deps key eg: check for something (existence of `composer.json` in `app.root`) and then add a `composer install` to `config.build`

dustinleblanc:
@serundeputy looks like this needs a rebase. Want to move it back to backlog or in progress?

PR #533, opened 2017-11-23
 Resolves #528 - Add intl and gettext extension for php 

### LINKED ISSUE #528: Add intl and gettext extension for php
Hey there,

first of all, great work! I really enjoy using lando, I absolutly love it! 

**Some suggestions regarding php:**
When you're working with translations, the php intl extension and gettext is often required. Currently it's really cumbersome to install this two extensions each time if I start a new project. The thing is, that the installation of intl takes it's time, because of it's dependeny to icu which needs to be compiled.

**Currently I'm doing this each time:**
services:
  appserver:
    extras:
      - "apt-get update -y"
      - "apt-get install libicu-dev  -y"
      - "docker-php-ext-install intl gettext"
      - "docker-php-ext-enable intl gettext"`

It would be much better, when those two extesion would be already installed. :)

benjaminhirsch:


benjaminhirsch:
Hmm I'm not really sure why the build is failing, respectively I'm not really familiar with the error. 🤷‍♂️  Maybe someone can help here? 😏

pirog:
@benjaminhirsch its failing on OSX during the step for setting up our certs to sign the macOS package. I think this is on us so ignore for now.

PR #547, opened 2017-11-28
Closes #545

### LINKED ISSUE #545: No Sudo No Cry
Make Lando un-sudo-able.

dustinleblanc:
This PR disallows running Lando from the CLI with `sudo` or with elevated privileges on Windows.

Running `sudo lando list` should cause an error message to be thrown and the process to exit. This doesn't work with just plain old `sudo lando` because of the `demandCommand()` call in cli.js. I've made a few attempts to override that and perhaps make use of yargs' `check` function, but I wasn't able to get that to work. We should consider adding more verbose output to give users more info on why this is bad, why they may be trying to do it, and how to fix that, perhaps just a link to docs.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -17,6 +17,7 @@ var chalk = require('./node').chalk;
 var os = require('os');
 var path = require('path');
 var Table = require('cli-table');
+var sudoBlock = require('sudo-block');
 
 // Yargonaut must come before yargs
 var yargonaut = require('yargonaut');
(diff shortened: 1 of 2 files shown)


dustinleblanc:
@pirog I started on tests for this too, but I couldn't quite think of the best way to test this using Chai, and I don't really know much about bats. Would you want tests? How would you want them implemented?

dustinleblanc:
Looks like windows needs elevated privs? https://ci.appveyor.com/project/pirog/lando/build/1.0.325#L1190

pirog:
@dustinleblanc windows does not need elevated privs except during install. im guessing appveyor just runs with elevated privs, at least at during that phase

dustinleblanc:
👍  Corrected the "" string

dustinleblanc:
@pirog Haven't figured out the magic incantations for Windows testing here. Should I maybe backup and go the no sudo -> Unix only route?

pirog:
@dustinleblanc yeah, maybe make this POSIX only for now and spin up a card for Windows? It's possible we don't even need this on windows because of how Samba handles the file sharing

Issue #551, opened 2017-11-28
Solr core not automatically created


hanoii:
I wanted to have, for now, just a solr service on a drupal project, not going full lando yet.

I have the following .yml file:

```yml
name: moog-drupal

proxy:
  # Lets get a nice lndo url for the solr web interface
  solrsearch2:
    - moog-solr.lndo.site:8983

# Spin up services to run a basic LAMP stack
services:
  solrsearch2:
    type: solr:6.6
    portforward: true
    core: drupal
    config:
      conf: .platform/solr-conf
```

For some reason the core is not created or found automatically.

Solr is there without core. I can see the configs on /opt/solr/server/solr/mycores/drupal, but the core is not there.

When I went to try to add the core manually, providing mycores/drupal as the instanceDir, it complained that it didn't have access to /opt/solr/server/solr/mycores/drupal/core.properties.

I ssh to the service and saw that /opt/solr/server/solr/mycores/drupal was owned by root:root. I then chowned it to solr:solr and added the core and that worked, but this surely have been automatically.

Using lando master.

heshanlk:
I have the same issue, the Solr core would not create automatically.

heshanlk:
After couple of different config attempts following configs appeared to be working fine for me for search api solr with Solr 5.5

```
  solr:
    type: solr:5.5
    portforward: true
    config:
      conf: docroot/modules/contrib/search_api_solr/solr-conf/5.x
```

heshanlk:
I was wrong, something this is working sometimes not so we need to fix for this.

hanoii:
@heshanlk you might want to check my pull request/fork and report back.

dmsmidt:
The PR of @hanoii fixes the problem.
This is currently keeping a complete company (20+ devs) from using Lando.

jbertoen:
I talked a lot in slack, a lot to myself as well, viva la rubberducking. @Mirnaxvb and @MikeWong1991  know what I mean.

The problem is the following:
- /solrconf is mapped to the conf dir as a volume by setting the conf value in the .lando.yml file
- /opt/solr/server/solr/mycores/CORENAME/conf is also mapped as a volume as a volume by setting the conf value in the .lando.yml file

But the thing is, when creating your solr service from scratch, without preexisting volumes or containers. The core needs to be created. Which would mostly mean that within /opt/solr/server/solr/mycores/CORENAME 2 things need to be added:
- core.properties
-data
[...]

PR #553, opened 2017-11-29
Do not map to the conf folder within the named volume

hanoii:
While trying to understand what was going on with #551, I noticed that there was an issue with permissions.

I found out by looking at the code that you are using a named mount possible to persist the solr data, and you are using the solr-precreate script. 

You were also mounting the custom config within the named solr volume, this causing a bit of a permissions issue as I noted on #551.

You really don't need to map the custom config within the named volume as that's taken care by the solr-precreate command.

### DIFF
--- a/plugins/lando-services/solr/solr.js
+++ b/plugins/lando-services/solr/solr.js
@@ -121,19 +121,12 @@ module.exports = function(lando) {
       solr.volumes = addConfig(globalConfig, solr.volumes);
 
       // If this is a recent version of solr we need to add to the config as an arg
-      // and also map to the core
       if (!_.includes(['3.6', 4.10], config.version)) {
 
         // Augment the start up command
         var command = solr.command.split(' ');
         command.push('/solrconf');
         solr.command = command.join(' ');
-
-        // Share the config to the core as well
-        var coreConf = path.join(solrConfig.dataDir, core, 'conf');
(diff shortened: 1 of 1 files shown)


hanoii:
@pirog/@reynoldsalec/@serundeputy should I be pinging you guys?

pirog:
@hanoii this actually seems to break things for me while the existing code works. Have you tried other versions? Would love to discuss this more so we can find a more generic solution for our supported SOLR versions that is also cross platform!

hanoii:
@pirog what does it break? It is odd that it works for you.. are you actually instantiating solr with a custom config? And what do you mean with versions, solr versions? I only tried 6.3 and 6.6.

jbertoen:
@pirog I have to agree with @hanoii here. I have been rebuilding for ages and the issue that I find is that the precreate-core script is not working in solr 5.5, 6.4 & 7.1.

After applying the suggested changes from @hanoii  the precreate script is working again.

dmsmidt:
This PR fixes the issue for me as well!

This is currently keeping a complete company (20+ devs) from using Lando.

pirog:
Alright all. I am checking this out now! 

Sorry for the delay and thanks for everyone's help to sort out the mess that i made!

pirog:
@hanoii

pirog:
closing this, manually merged into beta32 branch

PR #566, opened 2017-12-05
Doc-only: Correct "Accessing Services" link

tarnation:
Correct "Accessing Services" link

- [ x ] My code is just a doc update


pirog:
@tarnation nice catch thanks!

Issue #572, opened 2017-12-08
Add wkhtmltopdf to the install of tooling on the PHP images.


jbertoen:
Feature Request
===============

We often use WKHTMLTOPDF to generate our pdf files. Like imagemagick this works pretty good. Only the installation takes a lot of time because we have to do the following on every rebuild:
```
services:
  appserver:
    extras:
      - "apt-get update -y"
      - "apt-get install wkhtmltopdf -y"
      - "apt-get install xvfb -y"
```
The current docker file for php already implement similar logic for imagemagick and others. So why not add this as well.

pirog:
closing this with rationale in #573 
@jbertoen def feel free to open up a doc pr!

tanc:
I realise this is old but here is my successful install of `wkhtmltopdf` using the `drupal9` recipe with PHP 8.0:

```
services:
  appserver:
    build_as_root:
      - "[ -f /usr/bin/wkhtmltopdf ] || ( apt-get update -y && apt-get install zlib1g fontconfig libxrender1 libfreetype6 libxext6 libx11-6 -y)"
      - "[ -f /usr/bin/wkhtmltopdf ] || ( wget -nc https://github.com/wkhtmltopdf/packaging/releases/download/0.12.6-1/wkhtmltox_0.12.6-1.buster_amd64.deb -P /tmp )"
      - "[ -e /usr/bin/wkhtmltopdf ] || ( dpkg -i /tmp/wkhtmltox_0.12.6-1.buster_amd64.deb )"
      - "[ -f /usr/bin/wkhtmltopdf ] || ( ln -s /usr/local/bin/wkhtmltopdf /usr/bin/wkhtmltopdf )"
```

cbiggins:
@tanc this worked perfectly. Thanks.

yuseferi:
Nice!

melvix:
The solution posted by @tanc worked well for me until this week when I updated lando and changed my lando recipe to drupal10. The installation of wkhtmltox would fail because of a requirement of `libssl1.1`, which is apparently no longer supported.

Here is my workaround:

```
services:
  appserver:
    build_as_root:
      - "[ -f /usr/bin/wkhtmltopdf ] || ( wget -nc http://security.debian.org/debian-security/pool/updates/main/o/openssl/libssl1.1_1.1.1w-0+deb11u2_amd64.deb -P /tmp )"
[...]

PR #573, opened 2017-12-08
Add wkhtmltopdf to the install of tooling on the PHP images. #572

jbertoen:

improve our process.

### DIFF
--- a/plugins/lando-services/php/5.3-apache/Dockerfile
+++ b/plugins/lando-services/php/5.3-apache/Dockerfile
@@ -33,6 +33,8 @@ RUN apt-get update && apt-get install -y \
     xfonts-base \
     xfonts-75dpi \
     zlib1g-dev \
+    wkhtmltopdf \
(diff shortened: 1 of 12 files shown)


jbertoen:
@pirog I do not think CI broke on my PR. But on something else, something about gitbook.

pirog:
@jbertoen i def appreciate you submitting this PR but i dont think we are going to do this. The general rule that we *try* to enforce for our php images is that they should be as barebones as possible and users should use relevant lando mechanisms (eg build steps) to install additional dependencies. The main reason for this is that we can't install ALL THE THINGS in our php container.

The exceptions to this are `composer` since its pretty ubiquitous and can itself be used to grab more dependencies and common-ish php-extensions (and their underlying dependencies) since they are needed for most php frameworks and are generally pretty difficult and time consuming to install on the fly with lando build steps.

jbertoen:
@pirog I will create some documentation then.

PR #577, opened 2017-12-11
Add lando open command: lando open | lando open --url mysite.lndo.site

serundeputy:

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "lodash": "^4.15.0",
     "node-cache": "^4.1.1",
     "object-hash": "^1.1.8",
+    "openurl": "^1.1.1",
     "restler": "^3.4.0",
     "semver": "^5.4.1",
     "shell-escape": "https://github.com/thinktandem/node-shell-escape/tarball/master",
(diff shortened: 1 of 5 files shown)


pirog:
@serundeputy while i hate to support laziness this seems harmless enough ;)

one thing to note: if you are going to submit PR's like this it would be good to add in the other needed things to go along with it, eg docs and changelog entries. i'll add those now but lets hook it up in the futuretime

pirog:
@serundeputy looking at this a little deeper... the code for this seems potentially problematic. 

We are assuming the user has an appserver service and that that service has a proxy entry. Not sure we can reliably make those assumptions.

pirog:
@serundeputy i made some changes https://github.com/lando/lando/commit/c13b7fbff1013545619c956c87a042aa37fd69ac and merged this into the beta32 branch.

i tried to restructure it to be more lowest common denominator eg doesnt assume a service name or proxy setting but im sort of guessing this might still end up being more of a headache for developers since its far from certain `lando open` is going to actually going them the correct url. 

i sort of suspect we might end up removing this at some point because its just not giving users what they want and its currently not possible on our end to reliably give it to them it.

pirog:
@serundeputy yeah i tried this on a few examples and it just seems a little too finicky, even with the changes i made. We should discuss ways we could improve this so it doesnt add more trouble than its worth. When we get to that point we can just revert this commit to get back to a good place
https://github.com/lando/lando/commit/ae216980bf0a62ee576d5028d24e80a814b73d99

uberhacker:
Do you even need the `--url` flag?  Now I'm really talking like a lazy person. :)

PR #582, opened 2017-12-13
Fix travis macosx builds

serundeputy:


pirog:
@serundeputy are the `travis_wait`s still required?

serundeputy:
@pirog it's hard to say because at various times in the testing the timeouts were at different steps.

I think having them doesn't make it wait the whole time; just allows it to wait that long if necessary.

pirog:
@serundeputy im guessing we should probably remove them and see if it still works then

serundeputy:
@pirog build passed without the `travis_wait`s; thanks for clobbering the lingering `30`.

I think we are good to merge here.

PR #591, opened 2017-12-17
Add a nice lando logo on the default lando command for iterm users.

serundeputy:


serundeputy:
Add a lando logo for iTerm users for branding.

screen shot 2017-12-16 at 10 34 00 pm

pirog:
i think its super cool that this is possible but seems not really worth it?

Issue #362, opened 2017-09-25
Documentation on integrating lando docker setup with IDE like PHPStorm to run stuff like PHPUnit


jbertoen:
Documentation on how to integrate the lando docker setup as a PHP interpreter with IDE like PHPStorm to run tests for instance.

As a developer using lando and PHPStorm I would like them to be integrated so that I can run PHP Unit Tests and such from the IDE, to deliver better quality code in less time.

ptmkenny:
Yes, +1.

I will eventually try to get this working myself, but if someone already has it working and can just provide quick notes, I would be happy to clean it up into proper documentation.

In the meantime, there are other Docker projects with documentation on debugging/testing with PHPStorm [such as this on using XDebug](http://docs.docksal.io/en/latest/tools/xdebug/#debugging-with-phpstorm), and [this one on setting up the PHP interpreter](http://docs.docksal.io/en/latest/tools/behat/#integration-with-phpstorm).  Of course the steps aren't exactly the same, but for me it's been easier than starting with just the PHPStorm documentation alone, which is good but covers so many use cases it's easy to get confused.

jbertoen:
A tip for the people setting up xdebug without using xdebug=true in the lando config. Because this will slow down the site, because it will xdebug every request.
Use the following in your php.ini

```
xdebug.max_nesting_level = 256
xdebug.show_exception_trace = 0
xdebug.collect_params = 0
xdebug.profiler_enable_trigger = on
xdebug.remote_enable = on
xdebug.remote_host = ${LANDO_HOST_IP}
```

And something like this when setting up your appserver in .lando.yml
```
config:
  xdebug: false
  conf:
    php: php.ini
```

ptmkenny:
@jbertoen Could you explain the last line?

```
  conf:
    php: php.ini
```

This should point to a local php.ini with the xdebug configuration?

jbertoen:
@ptmkenny Exactly! This way you can pass your own php configuration, including xdebug. Using the above I got xdebug working. The only downside is that when I move from a physical location and my IP changes, I need to restart the containers for the project.

ptmkenny:
@jbertoen Ok, thanks.  I added a `php.ini` to my root directory and added the `conf:` line to `.lando.yml` and rebuilt my containers, but `phpinfo()` shows it was not picked up: the path is `/user/local/etc/php` or `/user/local/etc/php/conf.d/....`.

[...]

PR #600, opened 2017-12-22
Documentation on integrating lando docker setup with IDE like PHPStorm to run stuff like PHPUnit

Mirnaxvb:
Documentation on integrating lando docker setup with IDE like PHPStorm to run stuff like PHPUnit #362

improve our process.


pirog:
@Mirnaxvb this looks great! only thoughts i have is this might be better suited as a standalone "using lando with phpstorm" tutorial instead of embedded inside of the D8 one? What do you think?

Mirnaxvb:
Hi @pirog  would agree since you don't necessary need to use phpstorm for Drupal. I'm not sure how PHPUnit and stuff like that is set up if you use a other framework than Drupal. So i would suggest 2 options:

1. Create a section setting up PHPUnit for Drupal 8
2. Add some notes that the paths i used and the bootstrap part is for Drupal so you would get somthing like:
"As example with Drupal 8 we use the following paths...."

pirog:
@Mirnaxvb i'm doubtful there is a HUGE difference between drupal 8 and other frameworks so i'd just write a generic phpstorm tutorial and just use examples from Drupal 8... so i guess option 2!

anpolimus:
Tested at my local - this doc helped me to configure my phpstorm.
Thank you!

docs/tutorials/drupal8.md
```diff
@@ -217,3 +217,48 @@ Read More
...
 *   [Importing SQL databases](http://docs.devwithlando.io/tutorials/db-import.html)
 *   [Exporting SQL databases](http://docs.devwithlando.io/tutorials/db-export.html)
+
+### Setting up PHPStorm for running tests
+We want to be able to run PHPUnit, Kernel and Functional Test with PHPStorm instead by Commandline so we can debug our 
+tests. We want to be able to do this by the FPM from our lando app. To do this we need to setup a Remote interpreter 
+and make out Test Framework us it.
+
+* *Currently we have issues running functional test through PHPStorm. The reason this doesn't work yet is
+ that PHPStorm spins up a container and *
```
smk: And the reason is...?

dustinleblanc: trevor the vampire https://www.youtube.com/watch?v=mUj88_aPNkg

Issue #304, opened 2017-08-27
Add RAILz recipe


pirog:
Now that we've got a RUBY service dialed, the pathway to a RAILs recipe is open

@reynoldsalec maybe you can ping mrs @reynoldsalec to see if she knows of any useful documentation we could follow to make sure we cover all the rails basics.

serundeputy:
@dustinleblanc did you have some progress on a rails recipe?

dustinleblanc:
Yep! I got blocked though because for some reason, Puma was refusing connections. I set it up using vanilla docker-compose and Puma was happy to serve connections, so I may be getting blocked by a lack of Lando understanding. I can publish my branch on my fork if its helpful

serundeputy:
yes; I think publishing your branch will not hurt; and best case scenario encourage collaboration!

thanks for your great work so far!

dustinleblanc:
@serundeputy see https://github.com/lando/lando/pull/602. This is definitely not ready for prime time, but I got far enough that you guys may be able to correct my mistakes and we can work on rolling the ball further down the road.

hardyoyo:
Here's a Docker-Compose wrapper written in Ruby which is designed to set up a Rails dev environment, it might have some useful bits to swipe. https://gitlab.com/notch8/stack_car

dustinleblanc:
Marking this as a candidate for a new contrib plugin

PR #602, opened 2017-12-27
[WIP] Initial scaffold of Rails recipe

dustinleblanc:
per request from @serundeputy in #304 This is just a quick push of my WIP code for a Rails Recipe, it is definitely still funky. Some things I am battling:

- When you run `lando start` without the bundle built, Lando will keep running the stack while the Rails container is running `bundle install`, which can take a while. Since this isn't output to the screen, and can take forever, the check for the service being up is going wild on port 80 waiting for Puma to boot, but we haven't even brought the server up yet. I am fairly sure I saw @pirog mention something similar with node projects a while back in slack.
- Puma seems to dislike proxying the connection through traefik. I can't connect in a web browser, no matter what port I boot Puma on. I have googled and tried to find a solution but haven't been successful yet. This lead me to wonder if it would be better to run Puma via a socket, proxied behind Nginx, as that seems to be a good setup from what I read out on the webs (see https://www.justinweiss.com/articles/a-web-server-vs-an-app-server/). I may just be grabbing on to this because it sounded like a solution for the problem rather than being a better dev environment. I also dove into a deep hole of 'How does Heroku do it?' from which I haven't emerged and I have gotten busy with the holidays and other things.

To play with this, you need to publish a copy of the container somewhere your docker daemon can get to it, I had wierd intermittent issues with building it and then using it, but again, lack of workflow experience could be to blame. The only reason for a custom container is to add in some additional deps of a default rails app (node, pg and mysql connector clients).


dustinleblanc:
Additionally @pirog & @serundeputy, I copypasta'd a bunch of this from the Lamp recipe, we can probably pull out some mixins/modules/traits(whatevs JS calls them!) to make some things for recipe building more intuitive. Having a `canUsePg`, `canUseMySQL`, `canUseRedis`, and `canUserMemcached` or something would probably dry up recipe creation for all sorts of things. I am fairly sure, Django/Python, Express/Node, Rails/Sinatra/Ruby, allThePHPs, etc..., can connect to most of these services in a nearly identical way.

plugins/lando-recipes/rails/rails.js
```diff
@@ -0,0 +1,201 @@
+/**
+ * LAMP recipe builder
+ *
+ * @name lamp
```
hardyoyo: more refactoring required, s/b "rails" not "lamp", throughout

dustinleblanc: Absolutely true, though as indicated, this isn't intended to be done ;) There is far more work to do than that 😀

hardyoyo: Yup, I'm of the "start simple" school of thought. I'm working on a PR to your PR. I have no idea how to do this: "To play with this, you need to publish a copy of the container somewhere your docker daemon can get to it." But I'll dive in and see if I can figure it out.

hardyoyo: Oh, is this it? https://github.com/lando/lando/pull/602/files#diff-aad8a5410cee10fcbb10eb81fecd8762R3

dustinleblanc: yep! Exactly, that should work

ronald2wing:
@dustinleblanc Why stopped working on this issue? Do you have a rails recipe?

dustinleblanc:
@ronald2wing I just don't use Rails in everyday life, and I knew we were not likely to merge the recipe at the time.

That said, Lando has moved a LOT since I originally worked on this, and creating a rails app from scratch will Lando is ridiculously easy now. The Rails recipe is probably a really good candidate for one of the first first-party plugins to be supported on the new 3.1 plugin system we're building.

The from-scratch `.lando.yml` is less than 40 lines: https://github.com/dustinleblanc/lando-rails-example/blob/master/.lando.yml

this is a testament to how far Lando has come since I originally started hacking on this to teach myself how the plugin and recipe systems worked (and they have changed a bunch!)

Feel free to snag that landofile I linked above and let me know how well it works for you, I plan to write a guide about it some day

ronald2wing:
@dustinleblanc It is very helpful, I never know there is a compose type you can grab images from docker directly. I am interested if you have other useful landofile to share as well.

dustinleblanc:
Hey @ronald2wing. I guess it depends on what is considered useful :)

As specific issue comes up, I sometimes share new stuff in the Lando slack org, some times as gists, and I also have some repos on Github that are advanced Lando related that might be useful, depending on what you are doing.

ronald2wing:
@dustinleblanc Maybe you can add your rails Lando file in the Lando example.

PR #619, opened 2018-01-05
#618 Allow the usage of wildcards in proxy custom domain names for se…

### LINKED ISSUE #618: Allow the usage of wildcards in proxy custom domain names for services
Feature Request
===============

As a Lando user who works with Drupal/services/apps with multiple domains, I want an option to define my custom domains with wildcards to prevent adding a huge list of (sub) domains and rebuilding the app each time a new (sub) domain is needed. 

We have Drupal multi domain sites with about 100 defined domains, it is very tedious to maintain this list of domains at several places. It also makes starting lando very slow, since all domains will be checked every time the container starts.

We could solve this by allowing a wildcard symbol in the domains (e.g. *).

dmsmidt:
Fixes #618 

### DIFF
--- a/docs/config/proxy.md
+++ b/docs/config/proxy.md
@@ -62,6 +62,13 @@ Here is an example:
 127.0.0.1 billy.dee.williams
 ```
 
+Using Wildcards in Custom Domains
+---------------------------------
+
+If a service should be be able to listen to multiple domain names following a common pattern you can use the `*` wildcard character to match any amount of alphanumeric characters and hyphens/dashes (`-`).
+
+To match `site1.myapp.lndo.site` and `site2.myapp.lndo.site` you can for example use `*.myapp.lndo.site` or `*.*.lndo.site`.  
+
 Checking Proxy Routes
 ---------------------
 
--- a/plugins/lando-core/lib/scan.js
+++ b/plugins/lando-core/lib/scan.js
@@ -19,8 +19,10 @@ module.exports = function(lando) {
     // Scan our URLs and add a list of them and their status to the app
     app.events.on('post-start', 9, function() {
 
-      // Get app URLs
-      var urls = _.filter(_.flatMap(app.info, 'urls'), _.identity);
+      // Get non-wildcard app URLs
+      var urls = _.filter(_.flatMap(app.info, 'urls'), function(url) {
+        return _.identity(url) && url.indexOf('*') < 0;
+      });
 
       // Scan the urls
       return lando.utils.scanUrls(urls, {max: 17})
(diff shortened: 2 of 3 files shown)


pirog:
@dmsmidt how do i go about testing this? do i need to add the wildcarded domains to the proxy config or do you just get wildcard domains by default?

dmsmidt:
@pirog, I hoped it would be clear due to the updated docs and my comment in #618. So, when spinning up services wildcards don't work out of the box, that could give unexpected possible hits on certain URLs. So you really need to add a URL with (multiple) * to the proxy domain. I'll provide an example later if it's still not clear, let me know.

pirog:
@dmsmidt yup i thought it was clear but im asking because i tried to do exactly that eg:

```
proxy:
  web:
    - trivial-proxy.lndo.site
    - *.wildcardtest.lndo.site
```

And it threw an error so if you have a `.lando.yml` with `proxy` config that *should* work that would make it easier on my end.

dmsmidt:
@pirog, no computer with Lando with me, so will have to do that monday. Could you send me the error. Note that you should use quotes around the string with an asterisk.

pirog:
@dmsmidt it was a `yaml` parse error so i'm guessing it was the lack of quotes. I'll give it another try later this weekend. Sad we can't get it to work sans quotes since that is how all our other proxy things are documented but i guess we are just at the mercy of how yaml is parsed.

dmsmidt:
Yeah it is a pitty, the only workaround would be to use a reserved word, but that is fishy. All strings with special characters need to have quotes in yaml.

pirog:
@dmsmidt ok yeah that definitely works now. very cool. i'm going to note the quote requirement in the docs. it might also be worthwhile to surface these urls after a lando start so the user can see they are there.

dmsmidt:
@pirog Very nice!
Good idea to surface the URLs. 
Note that dubble quotes parse escape codes, single quotes suffice.

dmsmidt:
@pirog after your change I see that the wildcard domains are pinged again when surfacing.
This shouldn't be the case IMHO.

dmsmidt:
Snuck in a fix in #628

pirog:
I think they get sent to be scanned but they are automatically set as
resolved and then show up as yellow in the url list. They don’t actually
get scanned like the other urls.

dmsmidt:
@pirog, nope not true. It still gets scanned (see screenshot). The early return in #628 prevents that.
screen shot 2018-01-08 at 15 41 26

pirog:
@dmsmidt oh weird i thought `fulfill/reject` bumped you out of the promise so i guess we just want to `return fulfill()` instead

Issue #447, opened 2017-10-31
Feature request: Generate fully qualified vanity domain link with `lando drush uli`


uberhacker:
Lando currently requires `lando drush uli -l mysite.lndo.site` to generate a login link without `http://default/...`.  It would be nice if you could exclude the `-l mysite.lndo.site` part of the request if in the project directory.

uberhacker:
This issue appears to extend to Drupal multisite environments also.  It would be very helpful if the `-l ..` option could be omitted.

serundeputy:
@uberhacker 
  * Drupal 7: https://github.com/lando/lando/issues/339#issuecomment-329470443
  * Drupal 8: https://github.com/lando/lando/issues/339#issuecomment-329495496

uberhacker:
@serundeputy: I could do that but lando already knows my vanity domain, right?

pirog:
@uberhacker so the tricky part here is that lando can really only **make a good guess* about what URL should be used. The obvious issues are 

1. Vanity URLs are not universally used, or in some cases just don't work
2. The user could add (and prefer) a custom vanity URL 

There are various things we could do to mitigate those concerns but they seem excessive and still potentially brittle. Personally, i haven't been able to think of a good, simple and clever way to do this reliably. Def up for suggestions.

uberhacker:
You also have the situation where a request is made by https instead of http.  In that case, the hardcoded domain in settings.php causes issues.  Drupal internally already knows how to select the correct sites directory based on the url request so, in theory, multisite could "just work" :tm: as long as the project name matches a sites directory.

pirog:
Yeah i hear that although i think you can use use a protocol conditional based on drupals `drupal_is_https` or the `HTTPS` server var. Seen a bunch of people do that to setup dynamic protocol `$base_url`.

The issue is less on the Drupal side and more on the Lando config side eg we don't really have a universally reliable way to divine from Lando which URL the `drush` command should use

uberhacker:
Would this be over simplifying things?  What if you take the project `name: fubar` from `.lando.yml` and look in the sites directory for `fubar`.  If the directory exists, then use all the settings from that directory instead of default.

pirog:
@serundeputy this issue is plagued by the same underlying problem as the `lando open` command which is that we have no explicit notion of "primary URL" in `lando`, or a reliable way to programmatically assign one. 

[...]

PR #624, opened 2018-01-06
Adding lando open command.

serundeputy:


docs/cli/open.md
```diff
@@ -0,0 +1,19 @@
...
+
+Usage
+-----
+
+```bash
+# Opens a reasonable guess at your default url
+lando open
+
+# Opens a specific URL
+lando open --url https://myapp.lndo.site
```
uberhacker: Is the `--url` switch necessary?  I'm even lazier than you think. ;)

uberhacker: Right, but do you even need it with the override?  Can't you just use the domain as the argument?

serundeputy: Ahh; I see; you are probably right; I'll check

pirog: @serundeputy dropping a note here: might want to test this against non-recipes to get a sense of the issues we are up against

dustinleblanc:
Do we want to defer this till later or try to get it in this release?

pirog:
@dustinleblanc i think defer. Specifically i thin we need to figure out what to do re: https://github.com/lando/lando/issues/447#issuecomment-355786058 first

dustinleblanc:
Since we're deferring this, I am closing the PR

PR #628, opened 2018-01-08
#627 Fix scan url chalk for unresponsive urls and early return wildca…

### LINKED ISSUE #627: Chalk of surfaced urls which can't be pinged throw errors
Bug Report
==========

Using beta 33 on a mac.

The commit 5e1a50a7dc0f9ccb0cd7dbec2cf68013719d154d introduced an error during the surfacing of urls, if a url can't be reached. In that case the status is `false` and there is no color.

screen shot 2018-01-08 at 13 32 02

dmsmidt:
Fixes #627

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -120,8 +120,7 @@ exports.scanUrls = function(urls, opts) {
         // If URL contains a wildcard then immediately set fulfill with yellow
         // status
         if (_.includes(url, '*')) {
-          fulfill({url: url, status: true, color: 'yellow'});
-          return;
+          return fulfill({url: url, status: true, color: 'yellow'});
         }
 
         // Make the actual request, lets make sure self-signed certs are OK
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -117,9 +117,11 @@ exports.scanUrls = function(urls, opts) {
       // Send REST request.
       return new Promise(function(fulfill, reject) {
 
-        // If URL contains a wildcard then immediately set fufill with yellow status
+        // If URL contains a wildcard then immediately set fulfill with yellow
+        // status
         if (_.includes(url, '*')) {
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -162,7 +164,7 @@ exports.scanUrls = function(urls, opts) {
     .catch(function(err) {
       log.verbose('%s is not accessible', url);
       log.debug('%s not accessible with error', url, err.message);
-      return {url: url, status: false};
+      return {url: url, status: false, color: 'red'};
```
pirog: def + 1 here

lib/utils.js
```diff
@@ -117,9 +117,11 @@ exports.scanUrls = function(urls, opts) {
       // Send REST request.
       return new Promise(function(fulfill, reject) {
 
-        // If URL contains a wildcard then immediately set fufill with yellow status
+        // If URL contains a wildcard then immediately set fulfill with yellow
+        // status
         if (_.includes(url, '*')) {
           fulfill({url: url, status: true, color: 'yellow'});
+          return;
```
pirog: wondering if it's preferable (and if there is a difference) between this and just `return  fulfill({url: url, status: true, color: 'yellow'});`

pirog:
@dmsmidt looks good. added a small comment re `return` but otherwise thanks for cleaning up my mess ;)

dmsmidt:
@pirog I don't know bluebird, but fulfill has no return value, so the outcome is the same. Tested both.

pirog:
@dmsmidt let's roll this in then! Do we think this bug is big enough that we'd want to spin up a hotfix release?

dmsmidt:
@pirog it's up to you, my team switched to building Lando from source. I don't expect a lot of wildcard usage currently, and the error is only at the end of starting up when the url doesn't respond. App works fine though. So I think this can wait for the next beta.

PR #669, opened 2018-01-25
Make aversion of destruction the default action

derekderaps:
I don't think this code change actually works; just hoping to spark conversation about the merits of averting destruction by default (currently, pressing `<enter>` results in total descruction).

### DIFF
--- a/plugins/lando-core/lib/tasks/destroy.js
+++ b/plugins/lando-core/lib/tasks/destroy.js
@@ -19,7 +19,7 @@ module.exports = function(lando) {
       yes: {
         describe: 'Auto answer yes to prompts',
         alias: ['y'],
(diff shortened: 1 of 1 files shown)


pirog:
@derekderaps hmm yeah this seems not ideal although part of me feels like its an appropriate punishment for being too lazy to type in the yes/no.

pirog:
@derekderaps looks like the key here is you need to set the defaults for both the option parsing and the interactive questions

```js
      yes: {
        describe: 'Auto answer yes to prompts',
        alias: ['y'],
        default: false,
        boolean: true,
        interactive: {
          type: 'confirm',
          default: false,
          message: 'Are you sure you want to rebuild?'
        }
      }
```

PR #670, opened 2018-01-25
Removing reference (even in jest) to a racist slogan.

dzink:
I understand that it's a popular meme, and was probably meant to mock rather than support the ubiquitous racist slogan. Unfortunately it is also just one more unnecessary reminder of said slogan to a large number of POC (and allies) who could benefit from Lando and don't want to be reminded of it at every turn. (a bit of history of the slogan: https://www.huffingtonpost.com/esther-suh/the-racist-nostalgia-behind-make-america-trump_b_8145962.html )

It's probably time to retire it in favor of something less divisive. Feel free to pick a more light-hearted phrase that also gets you amped about Lando!

@pirog/@reynoldsalec/@serundeputy


pirog:
@dzink, to be clear, this is not being used as a meme or even in jest really. It is being used to openly mock our horrible horrible president and everything for which he stands. That said i understand the intention here is probably not obvious unless you personally know me and how i feel about 45 .

That said, maybe you could replace it with something less boring? Song lyrics are always fun.

PR #698, opened 2018-02-13
Add selfupdate command

uberhacker:

- [ ] I have pinged @pirog/@reynoldsalec/@serundeputy when I think my code is ready for prime time.

improve our process.

Tested and working on:
- [x] Linux
- [ ] Mac OS X
- [ ] Windows Pro


package.json
```diff
@@ -54,6 +54,7 @@
   "dependencies": {
     "bluebird": "^3.4.1",
     "cli-table": "^0.3.1",
+    "deasync": "^0.1.12",
```
dustinleblanc: I'm still learning the whole sync/async thing with node, but this *feels* like a different direction than the typical promisified direction we're taking in the rest of the code. I know that a lot of the first scripts I wrote weren't fully promisified. @pirog what do you think about that?

package.json
```diff
@@ -63,6 +64,7 @@
     "js-yaml": "^3.4.6",
     "jsonfile": "^2.4.0",
     "kalabox-stats-client": "^0.2.1",
+    "linus": "git+https://github.com/tomas/linus.git#v0.0.5",
```
dustinleblanc: This util looks handy! It has some pretty low usage numbers though. @pirog would you want to switch to something like this and add a dependency or just keep doing it ourselves?

uberhacker: I wish they had a sync method.  That would definitely help due to the current use case.

uberhacker:
@dustinleblanc: I appreciate the feedback.

uberhacker:
Closing due to merge conflicts and like an idiot, I pushed up to the master branch.

Issue #307, opened 2017-08-27
Add symfony recipe


pirog:

ro0NL:
Hi @pirog i did some initial work in user land :) https://github.com/ro0NL/symfony-lando

PR #699, opened 2018-02-13
Introduce Symfony 3.4 & 4.0 as per #307

legovaer:

improve our process.


legovaer:
@pirog I noticed that one of the steps has failed but I can't see why. I see that something went wrong during the deployment to platform.sh. 

Could you give me some more info so that I can reproduce the issue and fix it?

dustinleblanc:
@legovaer It looks to me like CI passed after you merged master back in, so all that's required is a review. Thanks for submitting a new Recipe! We're in the middle of some crazy refactor work at the moment before we move towards 3.0 stable, so I am not sure where we are on adding recipes, the triumvirate must convene to discuss, thanks!

serundeputy:
@legovaer this is looking good!

* I've re-rolled for latest master
* renamed some nginx.conv files to symfonyX.conf
* addressed some code standards (mostly we are using two space indentation)
* added in `base` var to allow users to use the `via` key for apache or nginx specification

Still needs a little work to get over the finish line.  I think right now it is giving forbidden, probably because I didn't get the config right on the `symfonyX.conf` files, but I do not want to hard code the `${LANDO_WEBROOT}/web` I think that ${LANDO_WEBROOT} should pick up what the user specified with `webroot` in their `.lando.yml` file.

@legovaer if you want to address these things that would be great!  Otherwise I'll try to get back to it this weekend.

quentint:
Hi, these recipes would be very helpful (I'm a newcomer): do you know when we'll be able to use them? Or can I use the files in https://github.com/lando/lando/tree/79cca956b5f43b5f7fd6e3022585c99c1c10dccd/plugins/lando-recipes/recipes/symfony34 and as a starting point?

Thanks for your help :)

PR #702, opened 2018-02-14
Document lamp appserver `via` option on recipe tut

acobster:
Just a small change to docs; no code changes.

This is already in the `lamp` example, but this is the example recipe that shows up on /tutorials/lamp.html, where I went looking for it.

### DIFF
--- a/examples/lamp2/.lando.yml
+++ b/examples/lamp2/.lando.yml
@@ -15,6 +15,10 @@ config:
   # NOTE: that this needs to be wrapped in quotes so that it is a string
   php: '5.6'
 
+  # Serve php by either apache or nginx. You can put in any `service:version`
+  # string that is supported by Lando's nginx and apache services.
+  via: nginx
+
   # Optionally specify the location of the webroot relative to your approot.
   #
   # If ommitted this will be your approot itself.


uberhacker:
@acobster: Doesn't this only apply to the `lemp` example?

acobster:
Doesn't look like it?

https://github.com/lando/lando/blob/master/plugins/lando-recipes/recipes/lamp/lamp.js#L81

And fwiw, the other lamp recipe has the exact same comment.

uberhacker:
I was just mentioning the *'a'* in lamp means apache.  The *'e'* in lemp means *e*nginx.

pirog:
LGTM!

pirog:
@uberhacker reviewing this again and @uberhacker is right. this config doesnt make any sense for `lamp` recipes because the `via` is always apache

acobster:
Yeah, LAMP and LEMP are definitely two different things[citation needed], it makes sense to enforce only one frontend on each. :)

I think my misunderstanding stemmed from [this block](https://github.com/lando/lando/blob/master/plugins/lando-recipes/recipes/lamp/lamp.js#L73) and the two that follow it. Those few blocks are pretty plainly written to support values other than `apache`. But looking more closely, I see that `via` is just hard-coded [here](https://github.com/lando/lando/blob/master/plugins/lando-recipes/recipes/lamp/lamp.js#L126), and the code in `resetConfig` is only there for recipes like `lemp` to use in extending `lamp`. Right?

That said, is there a reason to keep `via` in the existing [`lemp`](https://github.com/lando/lando/blob/master/examples/lemp/.lando.yml#L19) and [`lamp`](https://github.com/lando/lando/blob/master/examples/lamp/.lando.yml#L18) examples? I can see the value in keeping the supporting code there to make extending these recipes easy, but IMO suggesting an alternative `via` option in an example intended to support exactly one frontend is confusing (source: the existence of this thread).

pirog:
the `lemp` and `lamp` examples are meant to showcase and document how the `php` `service` works including the `via` option. they arent meant to support exactly one thing, they are two examples to demonstrate common stacks you could build with `services` straight up.

the `lemp2` and `lamp2` examples are meant to showcase `recipes` that use the `php` service and its `via` option behind the scenes to support exactly one mix of services codified in `recipe: lamp` or `recipe: lemp`.

PR #734, opened 2018-03-02
Closes #647 - Allow Remote Docker

### LINKED ISSUE #647: Remote docker support
I'm working on using lando on CircleCi Docker builder (not machine builder). They are using remote docker.

Currently, changing engine host is not possible since the config is not merged with values provided in `config.yml`.

https://github.com/lando/lando/blob/master/lib/daemon.js#L206

```
|exports.getEngineConfig = function() {

  // Create the default options
  var config = {
    host: '127.0.0.1',
    socketPath: '/var/run/docker.sock'
  };

  // Slight deviation on Windows due to npipe://
  if (process.platform === 'win32') {
    config.socketPath = '//./pipe/docker_engine';
  }

  // Return the config
  return config;

};
```

Question: Will allowing to specify custom host, port and ip, such as `tcp://35.227.77.180:2376` in `config.yml` connect to remotely started docker? Is this the right place for this change? Is providing host as  `tcp://35.227.77.180:2376` enough or Lando does more checks with socket etc?

[UPDATE]
CircleCI provides the following env vars after provisioning remote docker:
```
Allocating a remote Docker Engine

Assigned Docker Engine request id: 5482736
  provisioning: ..
Remote Docker engine created. Using VM 'prealloc-skuvyg9c-c586b8d4-dbb2-4086-b2a0-3051281ee476'
Created container accessible with:
  DOCKER_TLS_VERIFY=1
  DOCKER_HOST=tcp://35.196.9.70:2376
  DOCKER_CERT_PATH=/tmp/docker-certs138100934
  DOCKER_MACHINE_NAME=5482736
```

dustinleblanc:
This should allow folks to use Lando in environments like CircleCI where the Docker daemon is running on a remote server. The injected env var needs to be a JSON string for this to work. I can add docs too before we merge. @pirog thoughts on this? I wanted to add a test to this too, but pending the merge of some other test code, I didn't want to muck with duplicating stuff.

- [x] I have pinged [a project committer](https://docs.devwithlando.io/contrib/committer.html) when I think my code is ready for prime time.

### DIFF
--- a/bin/lando.js
+++ b/bin/lando.js
@@ -23,7 +23,7 @@ var userConfRoot = path.join(os.homedir(), '.lando');
 var version = require(path.join(__dirname, '..', 'package.json')).version;
 
 // Allow envvars to override a few core things
-var ENVPREFIX = process.env.LANDO_CORE_ENVPREFIX || 'LANDO_';
+var ENVPREFIX = process.env.LANDO_CORE_ENVPREFIX || 'LANDO';
 var LOGLEVELCONSOLE = process.env.LANDO_CORE_LOGLEVELCONSOLE || 'warn';
 var USERCONFROOT = process.env.LANDO_CORE_USERCONFROOT || userConfRoot;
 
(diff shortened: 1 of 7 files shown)


pirog:
@dustinleblanc just to keep a record: i think we decided in slack to figure out a generic way to handle keys that contain JSON objects. 

Not sure if you've made any progress since that convo but i'd be happy to look at/help on this after i blow through the rest of the issues!

PR #744, opened 2018-03-07
Add information for connecting IDE's to the Lando network.

KeyboardCowboy:
Provide instructions for integrating IDEs with the Lando network so your IDE can connect to services within Lando, such as bootstrapping a database when running Unit Tests.

- [x] I have pinged [a project committer](https://docs.devwithlando.io/contrib/committer.html) when I think my code is ready for prime time.

improve our process.


pirog:
@KeyboardCowboy definitely think this is good. that said the networking situation is going to change post beta37 so we might want to hold off on this and re-review once we get that out the door

dustinleblanc:
Beta 37 has dropped, we should re-review. @KeyboardCowboy  @pirog, lets make sure that we've properly captured current behavior

KeyboardCowboy:
@dustinleblanc @pirog I installed Beat 37 and tested the settings and the configuration I outlined still holds.  PhpStorm's default network setting of `bridge` does not allow it to communicate with the Lando network, but changing that setting to the Docker network name does.

PR #773, opened 2018-03-17
Created unit tests for the config module.

jmdrawneek:
Created unit tests for the config module.

- [N/A] My code includes documentation updates if relevant.


pirog:
Whole bunch of things to point out (and then probably document somewhere) here. I'll inline most of them but wanted to point out that there are already unit tests for this over at https://github.com/lando/lando/blob/master/test/unit/config.spec.js so you probably want to merge your work into there.

lib/config.js
```diff
@@ -1,18 +1,25 @@
 'use strict';
 
+/**
```
pirog: Any comment that starts with `/**` is going to get picked up by our JSDOC API doc parser and i don't think we want this getting picked up. I'd also just say leave this comment block out completely. People can read the name of the file so i don't think we are adding much here.

lib/config.js
```diff
@@ -1,18 +1,25 @@
...
 // Modules
 var _ = require('lodash');
 var fs = require('fs-extra');
 var os = require('os');
 var path = require('path');
 var yaml = require('js-yaml');
 
-/*
- * Document
+/**
```
pirog: @jmdrawneek ditto re `/**` here. I think we only want to use the double star comment blocks for exported functions

lib/config.js
```diff
@@ -24,25 +31,31 @@ var getSysConfRoot = function() {
 
 };
 
-/*
- * Determine whether we are in a browser or cli
+/**
```
pirog: @jmdrawneek ditto `/**` here. I think we only care about documenting/testing exported functions for now so i'd just leave "private" function alone for now. I'm guessing we will figure out exactly what we want to do with them down the road.

plugins/lando-init/tasks/init.js
```diff
@@ -138,6 +138,7 @@ module.exports = function(lando) {
         // Rebase on top of any existing yaml
         if (fs.existsSync(dest)) {
           var pec = lando.yaml.load(dest);
+          console.log(lando.utils.merger);
```
pirog: @jmdrawneek ? forgotten debug code?

pirog:
merged into and closing in favor of https://github.com/lando/lando/tree/772-unitTestConfig

Issue #795, opened 2018-03-20
Lando changes all the file permissions


neclimdul:
Bug Report
==========

Lando v3.0.0-beta-35 and source build on Ubuntu 17.10

`.lando.yml`
```
name: laraveltest
recipe: laravel
config:
  webroot: ./public/
  via: nginx
  php: '7.2'
```

Steps:
```bash
# Just to makes things fail a bit earlier and easier.
sudo rm -rf ~/.lando
lando start
```

Output:
```bash
# lando start
Recreating landoproxyhyperion5000gandalfedition_proxy_1 ... 
Recreating landoproxyhyperion5000gandalfedition_proxy_1 ... done
Creating network "laraveltest_default" with the default driver
Creating volume "laraveltest_appserver" with default driver
Creating volume "laraveltest_data" with default driver
Creating volume "laraveltest_data_database" with default driver
Creating laraveltest_appserver_1 ... 
Creating laraveltest_database_1 ... 
Creating laraveltest_database_1
Creating laraveltest_appserver_1 ... done
Creating laraveltest_nginx_1 ... 
Creating laraveltest_nginx_1 ... done
error:  Error: EACCES: permission denied, open '/home/neclimdul/.lando/cache/laraveltest:last_build'
    at Object.fs.openSync (fs.js:638:18)
    at Object.fs.openSync (pkg/prelude/bootstrap.js:466:32)
    at Object.fs.writeFileSync (fs.js:1287:33)
    at Object.writeFileSync (/snapshot/cli/node_modules/jsonfile/index.js:115:13)
    at Object.exports.set (/snapshot/cli/lib/cache.js:0:0)
    at AsyncEvents.<anonymous> (/snapshot/cli/plugins/lando-services/lib/config.js:0:0)
    at AsyncEvents.handle (/snapshot/cli/lib/events.js:0:0)
    at /snapshot/cli/lib/events.js:0:0
From previous event:
    at AsyncEvents.emit (/snapshot/cli/lib/events.js:0:0)
    at /snapshot/cli/lib/app.js:0:0
From previous event:
    at Object.exports.start (/snapshot/cli/lib/app.js:0:0)
    at /snapshot/cli/plugins/lando-core/lib/tasks/start.js:0:0
From previous event:
    at Object.run (/snapshot/cli/plugins/lando-core/lib/tasks/start.js:0:0)
From previous event:
    at /snapshot/cli/lib/tasks.js:0:0
    at runCallback (timers.js:800:20)
[...]

PR #805, opened 2018-03-23
#795: Fixed intermittent permission borking for linux users using alp…

pirog:
…ine services

improve our process.

### DIFF
--- a/docs/changelog/2018.md
+++ b/docs/changelog/2018.md
@@ -3,7 +3,8 @@ v3.0.0-beta.38
 
 **SERIOUSLY, READ THE DOCS!: https://docs.devwithlando.io/**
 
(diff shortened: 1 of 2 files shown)


pirog:
@dustinleblanc this is for #795 

Issue here is that we try to sync up the host/container user on linux. Usually we do this by syncing the host users ID/GID with `www-data` inside the container. That way both the host user and `www-data` can write files eg i can install a drupal module through the UI but also edit my code through sublime text. 

We had some alpine containers that do not have the required utils to do this mapping which causes the files to ultimately get owned by the container UID/GID (usually 82) in these scenarios.

So i've moved the code around a bit and added a guard to bail on the perm reset if the host/container situation doesnt match up

pirog:
@dustinleblanc thanks!

PR #813, opened 2018-03-24
659 systemctl

pirog:

improve our process.


pirog:
@dustinleblanc fucking windows

pirog:
@dustinleblanc check out the note over at https://github.com/lando/lando/pull/813/files#diff-6c30d1480cc5b876f5890e7b650c4103R99

I can confirm that upgrading to `shelljs@0.7.8` does allow this test to pass on windows as expected. I also did a cursory pass on whether other "things" (eg basic app start/stop/ssh) break from that upgrade and it **seems** to be ok on macOS/Linux/windows so **WE COULD** consider bumping the version, removing the test conditional for windows, and presumably getting the improvements in the shelljs bump. If we go that route we will need to push the relevant commits and probably get a little more testing to make sure nothing major is borked. 

Otherwise we can do the safe suggestion in the comment.

pirog:
@dustinleblanc yeah i think we should just fuck it and do it live. We've got a week of shakedown time using the bumped version before the new release and im guessing that should be sufficient. 

Let me know if you agree and i'll update this PR with the new versionz and things.

pirog:
@dustinleblanc alright alright alright, fixed and may the wind be on our backs!

Issue #511, opened 2017-11-16
Feature Request: Add Blackfire as a component of Lando 


tylerssn:
Since Blackfire does not add any overhead when not in use, it would be nice to have a parameter in our Lando config that toggled Blackfire. Something like: `blackfire: true`

Currently I am installing the Blackfire PHP extensions by extending the devwithlando/php image with:

    services:
      appserver:
        type: php:7.0
        overrides:
          services:
            image: my/lando-blackfire-php:7.0
        extras:
          ##########
          # Blackfire setup
          - "bash $LANDO_MOUNT/dev/lando/deploy/blackfire/init.sh"

The file located at `$LANDO_MOUNT/dev/lando/deploy/blackfire/init.sh` contains:

    # Configure Repo
    wget -O - https://packagecloud.io/gpg.key | apt-key add -
    echo "deb http://packages.blackfire.io/debian any main" | tee /etc/apt/sources.list.d/blackfire.list
    apt-get update
    
    # Install agent
    echo "Installing blackfire agent"
    apt-get -yqq install blackfire-agent
    echo -e "yes\n$BLACKFIRE_SERVER_ID\n$BLACKFIRE_SERVER_TOKEN" | blackfire-agent --register
    /etc/init.d/blackfire-agent restart
    
    # Install the Cli Tool
    echo -e "$BLACKFIRE_CLIENT_ID\n$BLACKFIRE_CLIENT_TOKEN" | blackfire config

I have a `.env` file that contains the `BLACKFIRE_*` variables referenced in the bash script above.

jbertoen:
Love to pick this up @serundeputy, if you are not already working on it.

jbertoen:
Some documentation from blackfire for reference: https://blackfire.io/docs/integrations/docker

dmsmidt:
Working on this!

jbertoen:
Hurry up or I'll take over @dmsmidt  😆 🤣

serundeputy:
@jbertoen @dmsmidt here is what I had going; https://github.com/serundeputy/lando/tree/511-blackfire if it helps

dmsmidt:
@serundeputy thanks, used some of your work.
I've got the integration working in https://github.com/dmsmidt/lando/tree/511-blackfire-support.

Still some things I want to work out. Like not using the entry-point script.. and maybe adding a blackfire agent cli tool `lando blackfire <url>`

dmsmidt:
[...]

PR #816, opened 2018-03-25
#511 blackfire support

dustinleblanc:
@pirog this should be good for review

I didn't add any tests for this one, we could consider a test strategy for services in the near future.


dustinleblanc:
WINDOWZZZZZZ

mmenozzi:
@dustinleblanc What's missing to have this merged?

dustinleblanc:
@mmenozzi It needs feature and/or unit tests

mmenozzi:
Would make sense to split this into two separate PRs? One for the blackfire agent service and another to have blackfire php extension installed on the php docker image. What do you think?

dustinleblanc:
Installation in the containers requires building a publishing new docker images as well. There really isn't a reason to split them up. Arguably installing the extension in all PHP services has likelylhood to cause issues and therefore needs tests. I'd go ahead and give this feature a +1, @awolfey also said he was interested, we should be able to get it wrapped up someday, we're all just quite busy. If someone wants to write the tests though I'd be happy to review when I have a minute

mmenozzi:
I could try to do this but it would be my first contribution to lando (at least for the js code) so don’t count on me. :)

dustinleblanc:
@pirog the tests I asked @mmenozzi to write require the pushed modifications to the dockerfiles to be in place to pass, this is a bit of a chicken/egg issue. How do you want to handle that kind of weirdness?

mmenozzi:
@dustinleblanc also the test which checks if the blackfire agent runs is failing but I think that is a race condition because even on my machine it fails the first time but then, when the container is fully started, it passes. How can we fix that?

dustinleblanc:
That's a good question @mmenozzi and I think a bigger one for this test suite as it grows. How do we prime that cache so that the docker images are always present? We also need to get around to fixing our testing configuration with Circle. I am still skeptical that the tests are running correctly

mmenozzi:
Ok @dustinleblanc I understand that test suite needs more work to be more reliable. Anyway I would track that work on a separate issue without blocking other feature development (like this). What do you think?

dustinleblanc:
See #983, we can track those fixes there. I think this branch is good, I just want to get @pirog's feedback on the test state given where we are at the moment.

javivf:
I'm trying to use this feature but I'm getting this error when I run `lando start:` 

`error: (HTTP code 500) server error - container ede42641442f9146a5f68bb35ba4f5cec45e9a8f9dd54653025b3660a9122b67 is not connected to the network lando_bridge_network`

any clue?

pirog:
@dustinleblanc im going to close this for now (but keep the branch alive) since we need to

1. Port to stable release (eg RC2+) plugin system
2. Get "lando plugin" helpers working

PR #828, opened 2018-03-28
788 unit test yaml

pirog:

improve our process.

### DIFF
--- a/bin/lando.js
+++ b/bin/lando.js
@@ -7,31 +7,32 @@
  * @name lando
  */
 
-'use strict';
+ 'use strict';
 
 // Modules
-var _ = require('lodash');
-var bootstrap = require('./../lib/bootstrap.js');
-var cli = require('./../lib/cli');
-var log;
-var metrics;
-var os = require('os');
-var path = require('path');
-var Promise = require('./../lib/promise');
-var sudoBlock = require('sudo-block');
-var userConfRoot = path.join(os.homedir(), '.lando');
-var version = require(path.join(__dirname, '..', 'package.json')).version;
+const _ = require('lodash');
+const bootstrap = require('./../lib/bootstrap.js');
+const cli = require('./../lib/cli');
+const os = require('os');
+const path = require('path');
+const Promise = require('./../lib/promise');
+const sudoBlock = require('sudo-block');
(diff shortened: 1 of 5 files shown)


bin/lando.js
```diff
@@ -7,95 +7,98 @@
...
   mode: 'cli',
   pluginDirs: [USERCONFROOT],
   userConfRoot: USERCONFROOT,
-  version: version
+  version
 };
 
 // Error handler
-var handleError = function(error, log, metrics) {
+const handleError = ({hide, message, stack, code}, log, metrics) => {
```
dustinleblanc: This looks like some object destructuring. I don't understand it much but I've avoided doing it thus far because of lack of understanding

test/unit/config.spec.js
```diff
@@ -17,11 +17,11 @@ chai.should();
 const config = require('./../../lib/config');
 
 // This is the file we are testing
-describe('config', function() {
+describe('config', () => {
```
dustinleblanc: We need to be careful with this, it can break Mocha, I've been trying to avoid it (hopefully I haven't committed any).

test/unit/yaml.spec.js
```diff
@@ -0,0 +1,103 @@
...
+const hasher = require('object-hash');
+const os = require('os');
+const path = require('path');
+chai.should();
+
+// Get yaml module to test
+const yaml = require('./../../lib/yaml')({error: () => {}});
+
+// This is the file we are testing
+describe('yaml', () => {
```
dustinleblanc: same issue, I think is one area where we have to avoid arrow functions because of lexical scope binding.

bin/lando.js
dustinleblanc: This isn't a blocker though, just a thought point

dustinleblanc: I've read about it and I think I get it now.

This just takes all the parts of `error` that we used in the code, and explictly pulls those properties into the function rather than making you say `error.message` or `error.stack`.

test/unit/config.spec.js
dustinleblanc: Since tests pass, I don't think we need to be strict

PR #827, opened 2018-03-28
Add pacman package support

uberhacker:
Fixes #814.

Thank you so much for contributing code to lando!

We will get to your PR as soon as we can but in the meantime you might want to
check to make sure you have done the following. **The more boxes you can check
the easier it will be for us to merge in your changes with confidence**

- [x] My code passes relevant CI status checks
- [ ] My code includes a test if applicable ([see here](https://docs.devwithlando.io/dev/testing.html))
- [x] My code includes an update to the `CHANGELOG` ([see here](https://github.com/kalabox/lando/tree/master/docs/changelog))
- [x] My code includes documentation updates if relevant.
- [ ] I have pinged [a project committer](https://docs.devwithlando.io/contrib/contributing.html#committers) when I think my code is ready for prime time.

If you are unclear about any of the above please add comments below so we can
improve our process.

Issue #814, opened 2018-03-24
Add support for arch


inverse:
# Feature Request
How difficult would it be to add support for arch? `fpm` should be able to make pacman packages right?

dustinleblanc:
@inverse part of the issue is testing/building releases on arch. The more platforms we support, the more we have to test.

I'd imagine to make the commitment to support arch based distros, we'd need to have a solid group of users wanting this.

anavarre:
Just registering I'd like this as well.

pirog:
@inverse @anavarre you can currently install on arch by just installing from source. it might be no big deal to roll pacman packages, and it might "just work", but the bigger issue is just providing support around that as @dustinleblanc mentions

inverse:
@dustinleblanc thanks for getting back to me. It definitely would be another dimension on the testing front.

@pirog That could be one thing we could do. I'm currently doing that but was thinking of a more user friendly way. Recently switched to arch and wanting to learn the packaging system in more detail. Creating a `lando-git` package on AUR to be a good way to dig into this topic.

pirog:
Summoning @uberhacker

uberhacker:
Lucky me.  See #827.

uberhacker:
Hey @inverse @anavarre @dustinleblanc @pirog: Please test out PR #827 when you get a chance.  Make sure you follow the steps outlined [here](https://docs.docker.com/install/linux/linux-postinstall/#manage-docker-as-a-non-root-user) and that you are testing on the master branch.  Earlier branches don't have `systemctl` support needed by Arch/Manjaro.  Also note if you are using Manjaro, you will need the AUR repository enabled for some of the dependencies needed by Lando.  Let me know how it goes!

pirog:
@uberhacker that info would be great to add to the install docs as part of the PR!!! ;)

pirog:
Going to close this out but there is EXPERIMENTAL/DEV support for this over here now
https://docs.devwithlando.io/installation/installing.html#development-builds

If the packages seem to be good then we can add them to our official releases.

inverse:
Thanks for pushing this out. I'll be running with this and will let you know!

pirog:
@inverse interesting. im guessing this is because its a dev release which is why you get that `-8` at the end of the version string. What's weird is the "dev" version should actually append something more like `-8-g8827`. This doesnt seem to be an issue with deb/rpm but maybe arch has stricter guidelines around version validation?

inverse:
[...]

PR #832, opened 2018-03-29
Add Arch/Manjaro specific package references and notes

uberhacker:
Merge this PR after https://github.com/lando/lando/pull/827.


ElijahLynn:
Looks like it is `sudo pacman --upgrade lando.pacman` per https://github.com/lando/lando/issues/814#issuecomment-376967802, which did work!

It isn't a normal thing to have a .pacman file, so we do need more instructions in the Linux install for Arch.

PR #841, opened 2018-03-30
785 unit test tasks

dustinleblanc:
Easiest test PR ever.

This module IMO could use some improvements. It doesn't validate the task object from what I can tell, there is only a single method and I don't think it needs much more coverage than this single spec for now.

### DIFF
--- a/lib/tasks.js
+++ b/lib/tasks.js
@@ -58,4 +58,12 @@ exports.tasks = [];
  * // Add the task to Lando
  * lando.tasks.add('destroy', task);
  */
-exports.add = (name, task) => this.tasks.push(_.merge(task, {name: name}));
+exports.add = (name, task) => {
+  const mergedTask = _.merge(task, {name: name});
+  // Minimal task validation, just checking that common properties exist. If not, throw error.
+  if (!(_.has(mergedTask, 'name') && _.has(mergedTask, 'command') && _.has(mergedTask, 'describe') && _.has(mergedTask, 'run'))) {
+    throw new Error('The task object is not valid, please see https://docs.devwithlando.io/api/api.html#landotasksadd');
+  }
+
+  return this.tasks.push(mergedTask);
+};
--- a/lib/tasks.js
+++ b/lib/tasks.js
@@ -1,7 +1,7 @@
 'use strict';
 
 // Modules
-var _ = require('lodash');
+const _ = require('lodash');
 
 /**
  * A singleton array that contains all the tasks that have been added.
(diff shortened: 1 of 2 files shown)


pirog:
@dustinleblanc yeah this USED to be much more complicated in kalabox before we moved to to yargs and simplified the task structure. 

+1 for validation but we probably want to do fairly weak validation eg check for the minimal set of required props for a task. Do you want to do that as part of this PR or an additional task?

dustinleblanc:
Yeah I can probably add a validator

pirog:
I think the minimal set for validation here are just `name`, `command`, `description` and `run`

dustinleblanc:
@pirog I am still working on the validation, decided to check this out: https://github.com/hapijs/joi

Seems like the most straight-forward way for us to do object validation project wide. Any objections to using this for the validation?

pirog:
@dustinleblanc i was thinking something a little less involved eg just checking to see if the task object has the given keys or something but i guess we can do this too.

dustinleblanc:
I'm not married to the deeper validation, if your spidey sense says err on the side of basic I can just check that the keys exist rather than validate the types of each key and marking some required, etc

pirog:
@dustinleblanc yeah i mean we've NEVER had an issue with task validation before so i dont think we need to do something that intense. Check for keys, if not there maybe throw an error with a link to the 
https://docs.devwithlando.io/api/api.html#landotasksadd or something like that.

Probably is fine to build it directly into the add function itself

dustinleblanc:
Validation simplified right into function, spec added, should be ready for review/merge

pirog:
@dustinleblanc very nice, always good to adhere to the glory of YAGNI

lib/tasks.js
```diff
@@ -58,12 +58,12 @@ exports.tasks = [];
...
-
-  // Build a task object
-  var tasker = _.merge(task, {name: name});
-
-  // Add to the task
-  this.tasks.push(tasker);
+exports.add = (name, task) => {
+  const mergedTask = _.merge(task, {name: name});
+  // Minimal task validation, just checking that common properties exist. If not, throw error.
+  if (!(_.has(mergedTask, 'name') && _.has(mergedTask, 'command') && _.has(mergedTask, 'describe') && _.has(mergedTask, 'run'))) {
```
pirog: @dustinleblanc not a block to merge but there's gotta be a ... shorter way to do this? 

```js
const requiredKeys = ['name', ... 'whatever'];
if (!_.isEmpty(_.difference(_.intersection(requiredKeys, _.keys(mergedTask), requiredKeys))
```
but better

dustinleblanc: Yeah I almost threw it into its own function, also wanted to method chain like
```js
_(mergedTask).has('foo').has('bar)
```
I'd actually prefer:

```js
if(!validTask(task))
```
And just do the dirty elsewhere...

dustinleblanc: Which actually, ruby has a convention of suffixing boolean functions with `?`

I'd absolutely love:
```js
unless validTask?(task) 
  throw a Fit()
```

Issue #626, opened 2018-01-08
Correct settings to XDebug in VSCode and Lando (drupal 8 recipe)


MatthieuScarset:
Hi Lando team, and thanks a lot for this great tool! :smile: 

I managed to run XDebug + Lando + Visual Studio Code (VSCode) so I'd like to share my configuration here hoping that it can help other dev.

Related issue: https://github.com/felixfbecker/vscode-php-debug/issues/222
Setting files: https://gist.github.com/MatthieuScarset/0c3860def9ff1f0b84e32f618c740655

Thanks!

pirog:
@MatthieuScarset oh very cool! Any ambition to turn this into something we can use in our docs?

MatthieuScarset:
Sure thing... can you please point me to the right place where I could add it?
thanks!

pirog:
I'd spin up a new markdown file in here https://github.com/lando/lando/tree/master/docs/tutorials and craft a tutorial about "using lando with vscode". Doesn't have to be comprehensive, just dump what you've done in there and hopefully over time we can incrementally build it up!

pirog:
@MatthieuScarset going to close this due to lack of activity but feel free to reopen if you submit a PR!

dawidnawrot:
I would add that opCache has to be disabled in order to make xdebug fuctional fully. OpCache cache some local variables and there variables are not accssible in IDE (whether it's a VSC or PHPStorm), so we can add an extra setting in php.ini file to disable it or even better disable opcache in settings.local.php file by providing: `ini_set('opcache.enable', 0);` expression.

PR #845, opened 2018-03-31
#626 - Add tutorial for XDebug in VS Code

MatthieuScarset:
- [ ] My code includes an update to the `CHANGELOG`

I have added a new page under tutorials to explain how to get XDebug working with Lando in VS Code. 

I would like to update the CHANGELOG but I was not sure how to edit this file actually.


pirog:
@MatthieuScarset this is looking good. I'd recommend adding an entry in https://github.com/lando/lando/blob/master/docs/SUMMARY.md before we merge so that the tutorial is visibile in the book hierarchy. There isn't a super obvious section it belongs in so i would drop it in "Advanced Usage". 

Let me know if you need any additional help on how to do all of that

Issue #853, opened 2018-04-01
Fundamental improvements to how database services set custom credentials and related commands


RobinHoutevelts:
**Tell us about your `.lando.yml`**

```yml
name: mysite
recipe: drupal8
services:
  database:
    creds:
      user: drupal
      password: drupal
      database: drupal
```

**Tell us about the command you were running**

```bash
lando init
lando start
lando db-import db.sql.gz
```

**Tell us about the error you got**
```
$ lando db-import db.sql.gz          
Preparing to import /app/db.sql.gz into drupal8 on database:3306 as drupal8...
ERROR 1045 (28000): Access denied for user 'drupal8'@'mysite_appserver_1.mysite_default' (using password: YES)
Destroying all current tables in drupal8... 
NOTE: See the --no-wipe flag to avoid this step!
Gunzipped file detected!
Importing /app/db.sql.gz...
ERROR 1045 (28000): Access denied for user 'drupal8'@'mysite_appserver_1.mysite_default' (using password: YES)
  96KiB 0:00:00 [16.1MiB/s] [>                                 ]  3%            

Import failed.
```

**Tell us generally about your bug**

I tried to import a database but got a `ERROR 1045 (28000): Access denied for user 'drupal8'@...` error.

While I can define `services.database.creds` in my `.lando.yml` file, it's values are not used by `db-import`.

Nor are those values put in the `appserver` `DB_*` environment variables.

**Tell us more**

I'm having trouble understanding how to override the database credentials that a recipe creates ( you can't , well not entirely).

People end up creating a `.lando.yml` file like I did. I have no idea where I came up with this, nor why I expected it to work, but it appears I'm not the only one (https://github.com/lando/lando/issues/843, https://github.com/lando/lando/issues/671, https://github.com/lando/lando/issues/593)

pirog:
@RobinHoutevelts Great write up here and even better explanation of what is going on. I think you correctly identified some other issues that this likely impacts (and im going to add another)

#843 #671 #853 #593 + #803 #727 

Reading through all these issues i think we should increase the scope of this ticket to address the following concerns

### 1. Poor Documentation

It is not clear at all what a user must do to change the database credentials once they've been set. We need to make sure this is documented in some sort of warning in all relevant services (and recipes depending on what we choose below). 

This is probably the most straightforward change we need to make and a no brainer. 

### 2. How do recipes/services interact wrt db (or any services) creds? 

[...]

Issue #843, opened 2018-03-30
Lando rebuild <database> is not awesome on drupal 7


jraviotta:
When configuring/debugging .lando.yml for a drupal 7 site, `lando rebuild` doesn't yield consistent results.

issues:

1. `<root>/sites/default/settings.php` is persistent and not updated with subsequent installs, nor mentioned in docs.
1.1 deletion of `settings.php` should be documented or handled by lando perhaps as a `settings.lando.php` 
2. `lando rebuild` is wonky when db credentials are updated in `.lando.yml`
2.1 Not sure what the problem is

<details>
<summary>lando config</summary><p>

```
{
  "servicesConfigDir": "C:\\Users\\jraviotta\\.lando\\services\\config",
  "servicesHelpersDir": "C:\\Users\\jraviotta\\.lando\\services\\helpers",
  "loadPassphraseProtectedKeys": false,
  "proxy": "ON",
  "proxyDomain": "lndo.site",
  "proxyDash": "58086",
  "proxyHttpPort": "80",
  "proxyHttpsPort": "443",
  "proxyHttpFallbacks": [
    "8000",
    "8080",
    "8888",
    "8008"
  ],
  "proxyHttpsFallbacks": [
    "444",
    "4433",
    "4444",
    "4443"
  ],
  "composeBin": "C:\\Program Files\\Docker\\Docker\\resources\\bin\\docker-compose.exe",
  "composeVersion": "3.2",
  "containerGlobalEnv": {
    "COLUMNS": 256,
    "LANDO": "ON",
    "LANDO_CONFIG_DIR": "C:\\Users\\jraviotta\\.lando",
    "LANDO_APP_NAME": "MySite",
    "LANDO_APP_ROOT": "C:\\Users\\jraviotta\\Documents\\Repos\\Drupal\\4pillars",
    "LANDO_APP_ROOT_BIND": "C:\\Users\\jraviotta\\Documents\\Repos\\Drupal\\4pillars",
    "LANDO_HOST_OS": "win32",
    "LANDO_HOST_UID": "1000",
    "LANDO_HOST_GID": "1000",
    "LANDO_HOST_IP": "host.docker.internal",
    "LANDO_WEBROOT_USER": "www-data",
    "LANDO_WEBROOT_GROUP": "www-data",
    "LANDO_WEBROOT_UID": "33",
    "LANDO_WEBROOT_GID": "33"
  },
[...]

Issue #671, opened 2018-01-28
Access denied for additional DB services (mysql, all?) causing db-import to fail.


Kolin:
Bug Report
==========

**Tell us about your setup**

v3.0.0-beta.35 on Windows 10 Professional

**Tell us about your `.lando.yml`**

```yml
name: llaia-website
recipe: drupal8
config:
  webroot: ./web

services:
  oldmembersdb:
      type: mysql
      creds:
        user: oldmembersdb
        password: oldmembersdb
        database: oldmembersdb
```

**lando info output:**

```yml
{                                                                                              
  "appserver": {                                                                               
    "type": "php",                                                                             
    "version": "7.1",                                                                          
    "via": "apache",                                                                           
    "webroot": "./web",                                                                        
    "config": {                                                                                
      "conf": "C:\\Users\\Kolin\\.lando\\services\\config\\drupal8\\php.ini"                   
    },                                                                                         
    "urls": [                                                                                  
      "https://localhost:32800",                                                               
      "http://localhost:32802",                                                                
      "http://llaia-website.lndo.site",                                                        
      "https://llaia-website.lndo.site"                                                        
    ]                                                                                          
  },                                                                                           
  "database": {                                                                                
    "type": "mysql",                                                                           
    "version": "latest",                                                                       
    "creds": {                                                                                 
      "user": "drupal8",                                                                       
      "password": "drupal8",                                                                   
      "database": "drupal8"                                                                    
    },                                                                                         
    "internal_connection": {                                                                   
      "host": "database",                                                                      
      "port": 3306                                                                             
    },                                                                                         
    "external_connection": {                                                                   
      "host": "localhost",                                                                     
      "port": "32801"                                                                          
    },                                                                                         
    "config": {                                                                                
      "confd": "C:\\Users\\Kolin\\.lando\\services\\config\\drupal8\\mysql"                    
    }                                                                                          
  },                                                                                           
  "oldmembersdb": {                                                                            
    "type": "mysql",                                                                           
    "version": "latest",                                                                       
    "creds": {                                                                                 
      "user": "oldmembersdb",                                                                  
      "password": "oldmembersdb",                                                              
      "database": "oldmembersdb"                                                               
    },                                                                                         
    "internal_connection": {                                                                   
      "host": "oldmembersdb",                                                                  
      "port": 3306                                                                             
    },                                                                                         
    "external_connection": {                                                                   
      "host": "localhost",                                                                     
      "port": "not forwarded"                                                                  
[...]

Issue #593, opened 2017-12-17
Adding second database container in a recipe doesn't respect `creds` key.


serundeputy:
Using the `backdrop` recipe with two mariadb databases I can not specify the `creds` for the second database.

```yml
name: civi
recipe: backdrop
config:
  webroot: web
  php: '7.0'

services:
  appserver:
    build:
      - "curl -sS https://platform.sh/cli/installer | php"
  database:
    type: mariadb
    portforward: 6618
  cividatabase:
    type: mariadb
    portforward: 6619
    creds:
      user: geoff
      password: candlabra
      database: cividatabase
```

The username is always `root` and blank `password` regardless of putting a `creds` key in your `.lando.yml` file

So you can still use the database you just have to know that the creds are `root` with an empty password.

I'm using the `backdrop` recipe here, but I suspect it is with any recipe and possibly `custom` as well (needs to be checked / tested).

serundeputy:
I tested this morning on a d8 recipe:

```yml
name: 593-d-8
recipe: drupal8
config:
  webroot: .

services:
  database:
    creds:
      user: geoff
      password: elaphone
      database: geoffalot
  mybase:
    type: mysql
    portforward: true
    creds:
      user: px
      password: px
      database: memow
```

The `creds` key is respected for the primary `database` container and I can login with the creds provided by `lando info`.

```json
{
  "appserver": {
    "type": "php",
    "version": "7.1",
    "via": "apache",
    "webroot": ".",
    "config": {
      "conf": "/Users/geoff/.lando/services/config/drupal8/php.ini"
    },
    "urls": [
      "https://localhost:32860",
      "http://localhost:32861",
      "http://593-d-8.lndo.site",
      "https://593-d-8.lndo.site"
    ]
  },
  "database": {
    "type": "mysql",
    "version": "latest",
    "creds": {
      "user": "geoff",
      "password": "elaphone",
      "database": "geoffalot"
    },
    "internal_connection": {
      "host": "database",
      "port": 3306
    },
    "external_connection": {
      "host": "localhost",
      "port": "32858"
    },
    "config": {
      "confd": "/Users/geoff/.lando/services/config/drupal8/mysql"
    }
  },
  "mybase": {
    "type": "mysql",
    "version": "latest",
    "creds": {
      "user": "px",
      "password": "px",
      "database": "memow"
    },
    "internal_connection": {
      "host": "mybase",
      "port": 3306
    },
    "external_connection": {
      "host": "localhost",
[...]

PR #854, opened 2018-04-01
Update `lamp` recipe to merge pre-configured services

RobinHoutevelts:

improve our process.

### DIFF
--- a/docs/changelog/2018.md
+++ b/docs/changelog/2018.md
@@ -3,6 +3,7 @@ v3.0.0-beta.40 - In development
 
 **SERIOUSLY, READ THE DOCS!: https://docs.devwithlando.io/**
 
+* Updated `lamp` recipe to merge pre-configured services [#853](https://github.com/lando/lando/issues/853)
 * Updated `terminus` to version `1.8.0` [848](https://github.com/lando/lando/issues/848)
 * Add Arch-based distro pacman package build support [#814](https://github.com/lando/lando/issues/814)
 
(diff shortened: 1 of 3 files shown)


RobinHoutevelts:
This PR fixes #853, #843, #671, #593 and probably a few other concerning overriding database credentials.

The main issue is that while the `lamp` recipe is responsible for creating the `DB_*` ENV variables on the `appserver`, it has no way of knowing if the user altered the `database` service in the `.lando.yml` file.

While the `mysql` service, _does_ pick up this change and creates the expected user and database.

This caused a lot of confusion.

RobinHoutevelts:
Basically it's now possible for a recipe to see if a user's `.lando.yml` already defined services. Before it could only see config.

And with that new possibility the `lamp` recipe can insert overridden `database.cred` into the ENV, fixing those issues.

@pirog If for some reason this is a no-no, you can close the PR as my understanding of lando is a bit better now and I have alternatives lined up.

pirog:
This is great and reading through the issue you posted actually makes me
more convinced this is a bug.

That all said, I think we got an opportunity here to resolve or move along
other impacted and/or related issues in one swift stroke by expanding the
scope of this a bit. I also think we want to change the implementation here
to work within the recipe->services control flow a bit better.

I’m AFK right now but I’ll try to pull in related issues with a set of
action items on your ticket when I’m back so we can move this great
improvement along

RobinHoutevelts:
@pirog Ok to close this PR since we've opted to go for more robust changes?

( Having `db-*` commands target the `database` service and deprecating `DB_*` env vars. https://github.com/lando/lando/issues/853#issuecomment-377873589)

pirog:
@RobinHoutevelts yes, good call!

PR #872, opened 2018-04-06
Updated Pantheon recipe to call out `git clone` pre-step

ccharlton:
Added bits that should clear up any confusion about `lando pull` not doing the initial `git clone ...` for users.

improve our process.

### DIFF
--- a/docs/tutorials/pantheon.md
+++ b/docs/tutorials/pantheon.md
@@ -56,7 +56,7 @@ Once you've started up your Pantheon site you will need to pull in your database
 
 ### 1. Using `lando pull`
 
-Lando provides a command for Pantheon sites called `lando pull` to get your database and files. **If you do not specify `--code`, `--database` or `--files` then `lando` will use the environment associated with your currently checked out `git branch`.** Please note, `lando pull` **does not** clone repos, it essentially does a `git pull` (merge) for convenience (see #2 above).
+Lando provides a command for Pantheon sites called `lando pull` to get your database and files. **If you do not specify `--code`, `--database` or `--files` then `lando` will use the environment associated with your currently checked out `git branch`.** Please note, `lando pull` **does not** clone repos, it does a `git pull` (merge) for convenience (see #2 above).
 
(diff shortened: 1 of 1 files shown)


ccharlton:
I forgot to update the line:

> # Initialize a Pantheon specific .lando.yml after getting code (`git clone ...`) from Pantheon

docs/tutorials/pantheon.md
```diff
@@ -55,7 +56,7 @@ Once you've started up your Pantheon site you will need to pull in your database
 
 ### 1. Using `lando pull`
 
-Lando provides a command for Pantheon sites called `lando pull` to get your database and files. **If you do not specify `--code`, `--database` or `--files` then `lando` will use the environment associated with your currently checked out `git branch`.**
+Lando provides a command for Pantheon sites called `lando pull` to get your database and files. **If you do not specify `--code`, `--database` or `--files` then `lando` will use the environment associated with your currently checked out `git branch`.** Please note, `lando pull` **does not** clone repos, it essentially does a `git pull` (merge) for convenience (see #2 above).
```
serundeputy: **does not** clone repos, it ~~essentially~~ does a `git pull` (merge)

@achton I think we should eliminate the word *essentially*

ccharlton: Yes, it is superfluous.

serundeputy:
thanks @ccharlton merged!

pirog:
@serundeputy @dustinleblanc i think we need to revert this merge and reconsider.

Ideally the document that is linked to at `Read up on how to get a .lando.yml` should contain all the information you need to get started regardless of where you start (eg on a pantheon recipe or elsewhere). That said it should be improved to include the git clone option and probably have its title and emphasis elevated. We definitely dont want to maintain separate specific instructions for each recipe that are common to all recipes. Also, the current three step process that is in there now is going to cause problems for people who `git clone` and then `lando init pantheon`.

So i'm proposing we 

1. Revert this merge, add in @ccharlton's good feedback to the generic getting started doc https://docs.devwithlando.io/started.html
2. Rename this step in our recipe "getting started" sections so its more obvious what its for and that its importance is emphasized.

PR #873, opened 2018-04-07
Revert "Updated Pantheon recipe to call out `git clone` pre-step"

pirog:
Reverts lando/lando#872

### DIFF
--- a/docs/tutorials/pantheon.md
+++ b/docs/tutorials/pantheon.md
@@ -13,10 +13,9 @@ You should also check out Pantheon's [local dev](https://pantheon.io/docs/local-
  Before you get started with this recipe we assume that you have:
 
  1. [Installed Lando](./../installation/system-requirements.md)
- 2. Your code is already cloned into the desired folder (`git clone ...`)
- 3. [Read up on how to get a `.lando.yml`](./../started.md)
+ 2. [Read up on how to get a `.lando.yml`](./../started.md)
 
- If after reading #3 above you are still unclear how to get started then try this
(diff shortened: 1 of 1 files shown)


pirog:
@serundeputy @dustinleblanc i think we need to revert this merge and reconsider.

Ideally the document that is linked to at `Read up on how to get a .lando.yml` should contain all the information you need to get started regardless of where you start (eg on a pantheon recipe or elsewhere). That said it should be improved to include the git clone option and probably have its title and emphasis elevated. We definitely dont want to maintain separate specific instructions for each recipe that are common to all recipes. Also, the current three step process that is in there now is going to cause problems for people who `git clone` and then `lando init pantheon`.

So i'm proposing we 

1. Revert this merge, add in @ccharlton's good feedback to the generic getting started doc https://docs.devwithlando.io/started.html
2. Rename this step in our recipe "getting started" sections so its more obvious what its for and that its importance is emphasized.

dustinleblanc:
I've read over the issue a few times and I am drawing a blank on what was changed and what is needed. To paraphrase what I am reading, it sounds like we're talking about the multiple entry paths to getting started:

1. You haz codez and you want to make them Lando-able
2. You haz nothing, you want to get the codez (like Kalabox Classic).

The commit we are undoing had assumed you needed to clone first, which isn't always true, so we want to generalize the docs to say, 

image

You can go with this, or you can go with that

Does that summarize what this is about? It sounds to me like the action item then is to define that generalized starting point?

ccharlton:
Need another copy edit for this? or we good?

PR #875, opened 2018-04-08
446/installers

dustinleblanc:
@pirog I think this does it for OSX/Linux, Do you know the magic for Windows?

### DIFF
--- a/installer/linux/scripts/postinst
+++ b/installer/linux/scripts/postinst
@@ -25,22 +25,16 @@ do
 done
 
 # Generate a Root CA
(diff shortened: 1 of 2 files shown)


pirog:
@dustinleblanc we will need to consult Dark Texts to discover the Forbidden Words that makes this work on Windows. 

This is great though. I've got the internal TLS stuff looking good for `beta.40` so if we can get this into `beta.41` we will have gotten rid of a major pain point.

dustinleblanc:
Closing this for now to cleanup. We'll win this later

PR #907, opened 2018-04-19
Fix minor issue with test output

JamesChevalier:
seems like there was a typo or misunderstanding
😆 🤣


JamesChevalier:
I suppose we should get buy-in from @serundeputy first. I'm not sure what your internal process is for these sorts of things. 😃

Issue #918, opened 2018-04-26
Documentation Issue with D7


wesruv:
On this page:
https://docs.devwithlando.io/tutorials/drupal7.html

It asks you to to setup the base_url:

```
// Set the base URL for the Drupal site.
$base_url = "http://mysite.lndo.site"
```

It should probably point out that you need to add the port number, I spent about an hour today trying to figure out why _every_ local asset was 404'ing ☹️ 

Everything else has been great, loving Lando!

PR #919, opened 2018-04-26
Adding port number to base_url instructions

wesruv:
Re: https://github.com/lando/lando/issues/918


dustinleblanc:
Thanks @wesruv! Doesn't hurt to be a bit more specific!

Issue #958, opened 2018-05-11
[Laravel] Inject APP_LOG => errorlog to send log output to stderr


dustinleblanc:
Laravel defaults to logging to a file, but wtf, we use Docker, all your logs are belong to stdout in docker land, so we should send the app logs here too

PR #965, opened 2018-05-16
#958 Inject App Log

dustinleblanc:

@pirog this should be all set, overridden the Pantheon way™

### DIFF
--- a/plugins/lando-recipes/recipes/laravel/laravel.js
+++ b/plugins/lando-recipes/recipes/laravel/laravel.js
@@ -47,6 +47,10 @@ module.exports = function(lando) {
     // Start by cheating
     var build = lando.recipes.build(name, base, config);
 
+    // Set the pantheon environment
+    var envPath = 'services.appserver.overrides.services.environment';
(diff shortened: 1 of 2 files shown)


plugins/lando-recipes/recipes/laravel/laravel.js
```diff
@@ -45,10 +45,14 @@ module.exports = function(lando) {
     config.php = _.get(config, 'php', '7.1');
 
     // Start by cheating
-    var build = lando.recipes.build(name, base, config);
+    const build = lando.recipes.build(name, base, config);
+
+    // Set the pantheon environment
+    const envPath = 'services.appserver.overrides.services.environment';
+    _.set(build, envPath, {APP_LOG: 'errorlog'});
```
pirog: i think doing this might be an issue @dustinleblanc, eg we are replacing the entire environment (which might already contain upstream envvars). Testing to verify whether that is true right now.

pirog: @dustinleblanc stand down! all clear here!

Issue #1024, opened 2018-06-09
Add multiple url's to the proxy for different ports on a single service


jbertoen:
Feature Request
===============

So you have a container (a node container for instance). You have 2 proxy url’s added:
```
proxy:
  node:
    - host1.localhost
    - host2.localhost:3000
```
This would use port 80 (because it is the first  url in the list) and not add host2.localhost
If I would put the 3000 version first, that one would be added to traefik, and the :80 would not.
And this is because you can not add “multiple rules” to a single service. They are merged in a single rule, if the ports match. Other url’s are skipped if they do not have the same port.

This piece of code in lando clearly generates a single set of url's for a single port per service.
```
        // Go through those services one by one
        .map(function(service) {

          // Get name port and hosts
          var port = _.get(service, 'routes[0].port', '80');
          var hosts = routes.stripPorts(service);
          if (port === 443) { port = 80; }

          // Add in relevant labels if we have hosts
          if (!_.isEmpty(hosts)) {

            // Get parse hosts
            var parsedHosts = routes.parseHosts(hosts);

            // Start building the augment
            return {
              name: service.name,
              service: {
                networks: {'lando_proxyedge': {}},
                labels: {
                  'traefik.docker.network': lando.config.proxyNet,
                  'traefik.frontend.rule': 'HostRegexp:' + parsedHosts,
                  'traefik.port': _.toString(port)
                }
              }
            };

          }
```
To improve this, my suggestion would be to make the labels multiple by adding sets per port

jbertoen:
This is ready for review btw!

PR #1027, opened 2018-06-10
Integrated the segment option for traefik in docker.

jbertoen:
This enables that you can add hosts for multiple ports per container/service.
Needed to update traefik as well, this contains fixes for segments in docker.

This would help solve issue #1024 .


pirog:
@jbertoen we are in the middle of a big testing push right now so going to live dangerously and just merge this in and see what happens!

PR #1060, opened 2018-06-26
resolves flow control issue by returning from promise.

### LINKED ISSUE #1059: Flow control is out of order for poweroff
When issuing a `lando poweroff` command, the proxy, and the message "Lando containers have been spun down." trigger prior to the actual containers being spun down.

image

This is caused by the `lando.app.stop(app);` call in poweroff.js not being returned to the promise, allowing the following code to run before the containers are stopped.

nullvariable:
fixes #1059


pirog:
OH YAH! LOOKING GOOD!

PR #1125, opened 2018-08-08
Add tinker command

hailwood:
One of the most common laravel commands is dropping into a tinker shell (Laravel REPL).
This PR simply adds the tool by default.

- [ ] <strike>My code includes a test if applicable</strike>
- [ ] <strike>My code includes documentation updates if relevant.</strike>

### DIFF
--- a/docs/changelog/2018.md
+++ b/docs/changelog/2018.md
@@ -6,6 +6,7 @@ If you are upgrading from pre-beta.40 follow the [beta.41 release note instructi
 **ALSO, STILL, SERIOUSLY, READ THE DOCS!: https://docs.devwithlando.io/**
 
 * Unit tested all of core [#978](https://github.com/lando/lando/issues/978)
+* Added `tinker` command to Laravel recipe [#1125](https://github.com/lando/lando/pull/1125)
 
 v3.0.0-beta.47 - [June 1, 2018](https://github.com/lando/lando/releases/tag/v3.0.0-beta.47)
 -------------------------------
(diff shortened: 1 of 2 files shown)


hailwood:
@dustinleblanc Pull request message states I should ping a committer, and there's no specialist for recipes, so you're the flavour of the day!

dustinleblanc:
@hailwood Thanks for the PR! While I agree that `php artisan tinker` is  a common Laravel command, we already provide `artisan` via `lando artisan` so I am not sure that short handing it to `lando tinker` makes sense, given a user can already `lando artisan tinker`. Later versions of Laravel also require you to pull in the tinker package, which would mean that this command would be broken for Laravel apps that don't include the package.

I'd like to get an opinion from @serundeputy on this, but I am inclined to say that we don't need to add this tooling entry. We'll let you know. Thanks!

serundeputy:
Thanks for the PR @hailwood !

I'm with you @dustinleblanc  if tinker is not in the Laravel core we can't count on it being present and working for the end user without installing the package for them.

I'd say this should be easy enough to add as a short command to your `tooling` section of `.lando.yml` something like:

```yml
tooling:
  tinker:
    service: appserver
    cmd:
      - artisan
      - tinker
```

So, easy enough to add yourself, but we can't put it in core, because we can't predict if the end user has installed the package.

dustinleblanc:
Thanks @serundeputy!

@hailwood thanks so much for the contribution. We're going to go ahead and pass on adding this to core, but if you have any other ideas for stuff that you'd find helpful for Laravel dev please throw up a new issue in the issue queue or jump on the slack org and chat with us about it. @serundeputy and I do Laravel work here at Tandem and are really interested in making Lando the best Laravel local dev solution for complex projects (you can't beat valet for simple ones!).

Issue #961, opened 2018-05-14
Custom services won't register when within the app. 


scotthooker:
So firstly my plugin is being picked up fine. But my service inside my plugin is not.

`  lando.events.on('post-bootstrap', function (lando) {
    lando.services.add('myservice', require('./lib/myservice')(lando));
  });
`

Despite this I consistently get

`warn: myservice is not a supported service.`

Are there any better docs or examples on implementing a custom service within a custom plugin?

scotthooker:
Ok so after chatting with @pirog on slack did some further playing/hacking/digging.

Custom services *do* work but only when defined within `~/.lando/`

Plugins which define a service or services within the app root do not get registered.

scotthooker:
Closing this for now. As lando is fine. :)

scotthooker:
Ok maybe since #765 was closed in favor of this. I'll reopen.

Essentially plugins can't live within the app. They need to live within .lando/

dustinleblanc:
@pirog is the plan of attack here to allow plugin loading within an app? If so, I think that's supa cool

tylerssn:
@dustinleblanc - as discussed in https://github.com/lando/lando/issues/597, I'd say that [Lando Plugin Anywhere](https://www.npmjs.com/package/lando-plugin-anywhere) is 90% there. If I can remember where I left off on that project, I think one of the outstanding issues I was working on was plugin dependency/loading order. Though, I think Lando's core events mostly made this feature negligible.

I'm happy to accept pull requests for anyone who wants to wrap this package up or simply hand the code over if the Lando team things the logic can be integrated into the core Lando package.

tylerssn:
@scotthooker - does this PR resolve the issue for you [#961](https://github.com/lando/lando/pull/1169)? 

I was unable to create recipes as project-level plugins for the same reason as you are reporting and this seems to have done the trick.

scotthooker:
This looks along the right lines! Will give it a go and test @tylerssn

danielrojas08:
I have a problem similar to this one. Here is what it says:
Unhandled rejection TypeError: Cannot read property 'builder' of undefined
    at Factory.get (/snapshot/lando/build/cli/lib/factory.js:0:0)
    at AsyncEvents.module.exports.app.events.on (/snapshot/lando/build/cli/plugins/lando-recipes/app.js:0:0)
    at AsyncEvents.handle (/snapshot/lando/build/cli/lib/events.js:0:0)
[...]

Issue #932, opened 2018-05-01
Plugin loaded, event registered but task is not added.


tylerssn:
Issue
=====

When I added a tasks via plugin to the plugins directory I expect to see the task listed when I execute `lando`. I do not see the task listed.

Details
==========

What is your lando version and operating system? Version: v3.0.0-beta.42 on MacBook 

```yml
#.lando.yml contains:
plugins:
  - task-local
```

Plugin `task-local` is within plugins directory.

```javascript
// plugins/task-local/index.js
module.exports = function (lando) {
    /**
     * Was throwing:
     *  Possible EventEmitter memory leak detected. 21 post-bootstrap listeners added.
     *  Use emitter.setMaxListeners() to increase limit
     */
    lando.events.setMaxListeners(lando.events.getMaxListeners() + 25);

    lando.events.on('post-bootstrap', function(lando) {

        // Define a task
        var task = {
            command: 'echo [appname]',
            describe: 'Test apples on [appname]',
            options: {
                yes: {
                    describe: 'Auto answer yes to prompts',
                    alias: ['y'],
                    default: false,
                    boolean: true,
                    interactive: {
                        type: 'confirm',
                        message: 'Are you sure you want to echo something?'
                    }
                }
            },
            run: function (options) {
                console.log(options);
            }
        };

// Add the task to Lando
        lando.tasks.add('apples', task);
    });
};
```

I can see that the plugin is registered with`lando -- -vvv`.

```bash
verbose: Loading event post-bootstrap priority 5
verbose: App moninamericas has compose files. /Users/tyler/.lando/compose/appname/appname-1.yml
verbose: Plugin task-local loaded from /apppath/plugins/task-local/index.js
info: App appname is ready!
debug: App appname has config name=appname, plugins=[task-local], 
```

Also, adding `console.log("output test")` to the `task-local/index.js` module will output `output test` when I execute `lando` but it does not output anything when placed within the `post-bootstrap` event.

scotthooker:
I'm having similar issues with a custom lando service. Basically won't register in time or looks like that post-bootstrap isn't doing anything.

pirog:
@tylerssn pinged you about this over on https://github.com/lando/lando/pull/1169. Hoping you can help us improve the plugin ecosystem!

Issue #765, opened 2018-03-16
Adding Lando Service via Plugin


lammensj:
Hi,

First, let me thank you for your AWESOME work on this amazing tool! It makes using Docker child's play.

Now back to the problem/question: I was trying to create a custom plugin which loads a custom service in order to use Elasticsearch 6.2 and Kibana. I can see the plugin being loaded in de logs (see line [139](https://gist.github.com/lammensj/91cb486ef2b1b2a287eefc2299bee0c6#file-lando-log-L139), but the event listeners are never handled, and therefor, my service never gets registered. This results in a [warning](https://gist.github.com/lammensj/91cb486ef2b1b2a287eefc2299bee0c6#file-lando-log-L119) saying that my service does not exists upon building.
I'm assuming it's somewhat related to [Adding Lando Recipe via Plugin](https://github.com/lando/lando/issues/597)

Here is a gist with all the code: https://gist.github.com/lammensj/91cb486ef2b1b2a287eefc2299bee0c6

Am I missing something?

pirog:
@lammensj going to close this in favor of #961

PR #1169, opened 2018-09-04
[Issue 961] - Make all events available to project-level plugins

tylerssn:
This PR resolves an issue with project-level plugins not being able to observe early events such as `post-bootstrap` (as explained in #961, #765 and, #932). This prevented services and recipes from registering properly when the plugin is installed at project scope.

Proof of concept:

1. `yarn add lando-plugin-magento-recipe`
2. Update project `.lando.yml`:

    ```yaml
    plugins:
      - lando-plugin-magento-recipe
    ```

- [X] My code passes relevant CI status checks
- [X] My code includes an update to the `CHANGELOG` ([see here](https://github.com/kalabox/lando/tree/master/docs/changelog))
- [X] I have pinged [a project committer](https://docs.devwithlando.io/contrib/contributing.html#committers) when I think my code is ready for prime time.


tylerssn:
@dustinleblanc - please review.

Etroid:
This is great, hope it makes it into the next version. Been testing it locally.

lib/cli.js
```diff
@@ -83,15 +84,27 @@ module.exports = class Cli {
...
     const srcRoot = path.resolve(__dirname, '..');
     const version = require(path.join(__dirname, '..', 'package.json')).version;
+    const configSources = [
+      path.join(srcRoot, 'config.yml'),
+      path.join(this.userConfRoot, 'config.yml'),
+    ];
+    const pluginDirs = [srcRoot, this.userConfRoot];
+
+    if (appRoot) {
+      configSources.push(path.join(appRoot, '.lando.yml'));
```
Etroid: I don't think you want to merge app specific `.lando.yml` settings with global lando `config.yml` settings. This will cause unexpected behavior.

Etroid: The only key we really need is `plugins`. Maybe we can try something like: 
```
const appConfigSources = [];
if (appRoot) {
      appConfigSources.push(path.join(appRoot, '.lando.yml'));
}
```

And then in [bootstrap.js L51](https://github.com/lando/lando/blob/master/lib/bootstrap.js#L51)
```
  // Include app specific plugins as well
  if (!_.isEmpty(config.appConfigSources)) {
    const appConfig = helpers.loadFiles(config.appConfigSources);
    const appPlugins = appConfig.plugins;
    config = helpers.merge(config.plugins, appPlugins);
  }
```

tylerssn: I have tested your suggestion locally and it worked out well. 

Just a note, the plugin suggested in my opening comment (`yarn add lando-plugin-magento-recipe`) needs an update pushed because it is no longer compatible with newer versions of Lando.

tylerssn:
@Etroid - the demo plugin `lando-plugin-magento-recipe` has been updated so that it is compatible with the latest Lando version. It can be used to test with this PR once more.

pirog:
@tylerssn im going to close this out because the plugin system is MUCH different now. That said, i think you are 100% correct about improving the ease of use and ecosystem around lando. Sadly we (wisely i think/hope) havent wanted to go too hard in this direction until we had a stable and locked down API.

When we cut RC2 we should have that so doing something "like this" is now on the table. What i would do, if you are still interested and willing is reach out to @dustinleblanc in the contibutors lando slack channel after the RC2 drops and see if you guys can draft up a proposal on how we can do this.

Issue #1143, opened 2018-08-20
Passphrase protected ssh key not being loaded


becizzz:
Bug Report
==========

**Tell us about your setup**

v3.0.0-beta17 on macOS High Sierra 10.13.6

**Tell us about your `.lando.yml`**

```yml
name: multisite
recipe: drupal8

compose:
  - compose.yml

config:
  webroot: web
  via: nginx
  php: '7.1'
  xdebug: true

tooling:
  build.sh:
    service: appserver
    description: Execute build.sh
    cmd:
      - ./build.sh
  codeception:
    service: appserver
    description: Run codeception
    cmd:
      - ./vendor/bin/codecept
  phpcs:
    service: appserver
    description: Run phpcs
    cmd:
      - ./vendor/bin/phpcs
  codesniff:
    service: appserver
    description: Run code sniffer with default config
    cmd:
      - ./vendor/bin/phpcs
      - "--standard=phpcs.xml"
  phpcbf:
    service: appserver
    description: Run phpcbf
    cmd:
      - ./vendor/bin/phpcbf
  codefix:
    service: appserver
    description: Run code autofix
    cmd:
      - ./vendor/bin/phpcbf
      - "--standard=phpcs.xml"
  theme-build:
    service: node
    description: Builds themes css files
    cmd: true
  sync.sh:
    service: appserver
    description: Execute sync.sh
    cmd:
      - ./sync.sh

services:  
  mailhog:
    type: mailhog
    hogfrom:
      - appserver
  
  appserver:
    ssl: true
    extras:
      - "/app/extras.sh"
    overrides:
      services:
        environment:
          WKV_SITE_ENV: lando

  elasticsearch:
    type: compose
    services:
      image: blacktop/elasticsearch:6.3
      command: /elastic-entrypoint.sh elasticsearch
      ports:
        - "9200:9200"

  kibana:
    type: compose
    services:
      image: blacktop/kibana:6.3
      environment:
        KIBANA_ELASTICSEARCH_URL: http://elasticsearch:9200
      ports:
      - "5601:5601"
      command: /docker-entrypoint.sh kibana

  node:
    type: node
    run:
      - cd $LANDO_MOUNT/web/themes/custom/defender/ && yarn install

  db_1:
    type: mariadb:10.1
    portforward: 32001
  db_2:
    type: mariadb:10.1
    portforward: 32002
  db_3:
    type: mariadb:10.1
    portforward: 32003
  db_4:
    type: mariadb:10.1
[...]

PR #1172, opened 2018-09-06
Updates docs for passphrase protected SSH keys

tanc:
A documentation change for the problem identified in #1143


dustinleblanc:
Thanks @tanc! Going to take a crack at confirming this with our new functional testing framework!

PR #1184, opened 2018-09-10
Include note on how to run the functional tests

dustinleblanc:


pirog:
@dustinleblanc lol yes, good call

PR #1211, opened 2018-09-25
Simplify chown/chmod permissions in user-perms.sh

### LINKED ISSUE #1197: [v3.0.0-rc.1] Drush not found in Drupal recipes
Bug Report
==========
```bash
$ lando drush
/bin/sh: 1: drush: not found
```
It appears that the home directory for the normal user has changed from `/var/www` to `/user` and when you ssh into the container and cd to the home directory, it is still `/var/www`.  This wouldn't normally be a problem but since drush has been installed in `/user/.config` instead of `/var/www/.config`, the path to the drush executable cannot be found.  See below:
```bash
$ lando ssh -u root
root@576ca18fb27f:/app# find / -name drush | grep bin
/user/usr/local/bin/drush
/user/.config/composer/vendor/bin/drush
root@576ca18fb27f:/app# echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/var/www/.composer/vendor/bin
```
```bash
$ lando ssh
www-data@576ca18fb27f:/app$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/var/www/.composer/vendor/bin
www-data@576ca18fb27f:/app$ cd
www-data@576ca18fb27f:~$ pwd
/var/www
```
System specs:
```bash
$ uname -a
Linux Lenovo-Y70-70-Touch 4.15.0-33-generic #36~16.04.1-Ubuntu SMP Wed Aug 15 17:21:05 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID:	LinuxMint
Description:	Linux Mint 18.2 Sonya
Release:	18.2
Codename:	sonya
$ lando version
v3.0.0-rc.1
$ docker -v
Docker version 18.06.1-ce, build e68fc7a
$ docker-compose -v
docker-compose version 1.22.0, build f46880fe
```
Contents of `.lando.yml`:
```yml
name: mysite
recipe: drupal7
config:
  webroot: .
services:
  database:
    type: mysql:5.6
  appserver:
    type: php:5.6
    via: apache:2.2
```
Symptoms when executing `lando start` for the first time or `lando rebuild`:
```bash
...
Creating mysite_appserver_1 ... done
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   599    0   599    0     0   1693      0 --:--:-- --:--:-- --:--:--  1696
100 4482k  100 4482k    0     0  1842k      0  0:00:02  0:00:02 --:--:-- 2716k
 PHP configuration      :                  
 PHP OS                 :  Linux           
 Drush script           :  /tmp/drush.phar 
 Drush version          :  8.1.16          
 Drush temp directory   :  /tmp            
 Drush configuration    :                  
 Drush alias files      :                  

mv: cannot create regular file '/usr/local/bin/drush': Permission denied
error: Looks like one of your build steps failed with [object Object]
warn: This **MAY** prevent your app from working
warn: Check for errors above, fix them, and try again
```

uberhacker:
Fixes #1197.

- [x] My code includes an update to the [CHANGELOG](https://github.com/kalabox/lando/tree/master/docs/changelog)

If you are unclear about any of the above please add comments below so we can improve our process.

### DIFF
--- a/docs/changelog/2018.md
+++ b/docs/changelog/2018.md
@@ -1,3 +1,5 @@
+* Simplify chown/chmod permissions in user-perms.sh [#1197](https://github.com/lando/lando/issues/1197)
+
 v3.0.0-rc.1 - [September 11, 2018](https://github.com/lando/lando/releases/tag/v3.0.0-rc.1)
 ----------------------------------
 
--- a/plugins/lando-engine/scripts/user-perms.sh
+++ b/plugins/lando-engine/scripts/user-perms.sh
@@ -65,27 +65,11 @@ if [ $(id -u) = 0 ]; then
 
   fi
 
-  # Make sure we set the ownership of the mount and HOME when we start a service
-  echo "And here. we. go."
-  chown $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /var/www
-  chown $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /usr/local/bin
-  chmod 755 /var/www
-  chown $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP "$LANDO_MOUNT"
-
-  # Do a background sweep
-  nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /var/www/.ssh &>/dev/null &
-  nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /var/www &>/dev/null &
-  nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /usr/local/bin &>/dev/null &
-  nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP "$LANDO_MOUNT" &>/dev/null &
-  nohup chmod -R 755 /var/www &>/dev/null &
-
-  # Make sure we chown the $LANDO_WEBROOT_USER home directory
-  nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP $(getent passwd $LANDO_WEBROOT_USER | cut -d : -f 6) &>/dev/null &
-  nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /user/.ssh &>/dev/null &
-  nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /lando &>/dev/null &
-
-  # Lets also make some /usr/locals chowned
-  nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /usr/local/lib &>/dev/null &
-  nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /usr/local/share &>/dev/null &
-
+  # Make sure we chown the appropriate directories
+  chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP "$(getent passwd $LANDO_WEBROOT_USER | cut -d : -f 6)" &>/dev/null
+  chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP "$LANDO_MOUNT" &>/dev/null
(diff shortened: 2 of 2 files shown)


pirog:
@uberhacker i definitely appreciate the simplicity but i think there are a few other things to consider here. Guessing it won't take a modify this to take those things into account.

Basically what we need here is a group of blocking and non-blocking `chmod/chown`s. AFAIK these are blocking operations so if someone has a HUGE home folder it might actually take way to long to wait for these things.

I think we want to group them like this, 

```bash
# Non-blocking chmods on things that need to be set but we can wait on
nohup chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /some/non-critical/or/potentially/large/dir &>/dev/null &


# Blocking things that MUST be set before we do other stuff eg build steps
chown -R $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /usr/local/bin &>/dev/null 
```

The big thing here is figuring out which paths are "critical" and which are not. The old code assumes the following are the critical ones. What do we need to add here to satisify #1197 

```bash
# "critical paths" based on the current code
chown $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /var/www
chown $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP /usr/local/bin
chmod 755 /var/www
chown $LANDO_WEBROOT_USER:$LANDO_WEBROOT_GROUP "$LANDO_MOUNT"
```

uberhacker:
@pirog: The recursive chown is almost instantaneous with 4+ gb on my system (like less than 2 seconds).  On the other hand, things start to slow down considerably when recursively chmod-ing even if I process with the find command and isolate directories.  That is beside the point, however.  What we don't want to do is process the perm fixes asynchronously instead of synchronously.  By prefixing the command with `nohup`, essentially that is what you are doing.  I imagine just removing that with the existing code will fix issue #1197.  Can you please summarize the difference between the `/user` and `/var/www` directories?  I'm not understanding the purpose of having 2 _"home"_ directories.

pirog:
Removing `nohup` for *all* the directories is not an option. While it might not be an issue on your machine that is undoubtedly going to be prohibitively slow on other machines and more importantly if we don't need to run some directories synchronously then we shouldn't. **WE ONLY WANT TO SYNCHRONOUSLY PERM FIX THE DIRECTORIES THAT ARE CRITICAL**

`/user` is a mount of your host home directory
`/var/www` is the home directory for the container user

uberhacker:
There is no harm in recursively chown-ing the entire `/usr/local` directory and subdirectories in the container and like I said, chown-ing takes no time to process.  I guess we can leave the `/user` directory alone then.  There should be no reason to ever chown/chmod anything since it won't affect container operations.

pirog:
@uberhacker im definitely open to synchronously and recursively chowning all of `/usr/local` since that seems like the potential problem area and i'm pretty sure we can guarantee this directory is going to be pretty "light" in all our containers. It's the other, potentially larger, directories that you've switched to be synchronous that are problematic 

I also haven't had a chance to actually replicate this issue on linux and then see if this fixes things, this would be a great candidate for a new functional test as well.

uberhacker:
The only directory we are recursively chmod-ing is `/var/www` (the only potentially slow operation) and you currently do that without the nohup prefix, so I guess I'm not understanding why removing nohup is not an option?

pirog:
@uberhacker `chown` on a large directory is a potentially slow operation. Removing `nohup` is not an option for *all* the directories eg what you've proposed in the PR. Removing `nohup` for problematic directories eg `/usr/local` is definitely possible.

uberhacker:
Why do you need to chown/chmod on any directory that is not problematic?  Sorry, playing the devil's advocate here. :-P

pirog:
@uberhacker All the directories we are modifying here are potentially problematic. The key difference, and the difference that determines whether we sync or async is whether we need to "fix" those problems immediately because some other build step needs things a certain or whether we can wait.

i'd encourage you to read any of the innumerable articles about doing container/host permission syncing when doing local dev with Docker on linux.

Issue #1145, opened 2018-08-21
Make publicly available url more customizeable


osopolar:
Would be nice if the `lando share` url (the local tunnel) could be set differently to <projectname>.localtunnel.me with an option like --hostname example-a82b8917 for http://example-a82b8917.localtunnel.me.

PR #1219, opened 2018-09-28
Fix lando share command.

serundeputy:


examples/share/README.md
```diff
@@ -0,0 +1,52 @@
...
+=============================
+
+This example provides tests for the `lando share` command.
+
+Getting Started
+---------------
+
+This example is here for functional testing only.
+
+Helpful Commands
```
serundeputy: @pirog @dustinleblanc for some reason when the parser creates the `share.spec.js` file it is getting the tests to run from this `Helpful Commands` section and not from the `Testing` section.

examples/share/README.md
```diff
@@ -0,0 +1,52 @@
...
+----------------
+
+```bash
+# Get app info
+lando info
+
+# Verify the lando share command works
+lando share --url http://localhost:8081
+lando ssh -c "curl -I https://share.localtunnel.me |grep \'200 OK\'"
```
dustinleblanc: This is going to be quite brittle, you can get a different domain assigned from localtunnel depending on what else is using it. parallel test runs or any other person using that domain might collide

examples/share/www/index.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+phpinfo();
```
dustinleblanc: we might want to echo/print a specific string that we can grep for later

plugins/lando-app/tasks/share.js
```diff
@@ -49,7 +49,7 @@ module.exports = lando => {
           // Get the URLS
           .then(() => {
             // Assume a port to start
-            const port = 80;
+            var port = 80;
```
dustinleblanc: we should use let instead of var here.

examples/share/.lando.yml
```diff
@@ -0,0 +1,81 @@
...
+    #   drush/drush: "~8"
+
+  appserver:
+    overrides:
+      services:
+        ports:
+          - '8081:80'
+
+  # My database service
+  database:
```
dustinleblanc: we can probably make this test faster by not adding extra services. A bare minimum apache/nginx, or even node service that can boot fast and just shoot back a response would speed up the tests and be lower bandwidth

examples/share/.lando.yml
```diff
@@ -0,0 +1,81 @@
...
+    type: mysql
+    portforward: 3308
+    creds:
+      user: lamp
+      password: lamp
+      database: lamp
+
+# Add some nice command routing
+tooling:
```
dustinleblanc: there are probably not relavent to the test case, we can truncate these out

examples/share/README.md
dustinleblanc: parser might need some tweaks

pirog: hmmm @serundeputy we've got other tests that have a "helpful commands" section so id look at those and compare. here is an example
https://github.com/lando/lando/tree/master/examples/python

examples/share/README.md
pirog: i dont think we have an option to pass in a custom domain to localtunnel so either we need to 

1. Implement that option as part of this (https://github.com/localtunnel/localtunnel#opts)
2. Change the app name from share to something sufficiently obscure so as to reduce the brittleness

examples/share/.lando.yml
pirog: +1

examples/share/.lando.yml
pirog: 👍 good suggestion

plugins/lando-app/tasks/share.js
pirog: yes, `let` is now the preferred syntax in our es6 things

pirog:
@serundeputy bonus challenge, get this code A rated on code climate: https://codeclimate.com/github/lando/lando/plugins/lando-app/tasks/share.js

im guessing we probably wont have another dedicated sprint to up the codeclimate score like we did before we improving things as we go is probably the way to go

examples/share/.lando.yml
serundeputy: The problem i'm running into here is that the base services `apache` and `nginx` do not have curl installed on the base image.  This test depends on making a curl request.

The simplest thing I've found so far that has curl is `lamp` and `lemp` recipes.

I could add this test to one of those recipes ( that is what I initially did), but it also requires specifying a fixed port and I did not necessarily want to introduce that into those existing tests ... although we could.

pirog: @serundeputy ahhhh right, that makes sense to me. i'd just roll with the `lamp` stack then since it's one fewer containers than `lemp`

examples/share/README.md
pirog: @serundeputy looks like there is already a card for the custom domain option: lando/lando#1145

pirog:
@serundeputy i got this merged into the leia branch so closing it out here

PR #1241, opened 2018-10-10
Job opportunity notification

reynoldsalec:
@pirog here's the docs modifications that I'm suggesting to notify people about the open gig.


pirog:
@reynoldsalec LGTM, i think at some point its likely we will have markdown standards so i modified to be in line with what i assume they will be

Issue #1086, opened 2018-07-06
MongoDB service doesn't work


hussainweb:
Bug Report
==========

**Tell us about your setup**

What is your lando version and operating system? **(note that the older your version the less likely we are to reply)**

v3.0.0-beta47 on MacOS High Sierra

**Tell us about your `.lando.yml`**

```yml
name: laravel-app
recipe: laravel

services:
  database:
    type: mongo:3.4

config:
  php: 7.2
  webroot: public
  xdebug: true

tooling:
  mongo:
    service: database
    description: Drop into the mongo shell
```

**Tell us about the command you were running**

Paste in a series of commands that caused the bug.

```bash
lando start
```

**Tell us about the error you got**

Rerun the command that produced your error in `verbose` mode. Paste in the error and the ~100 lines that proceeded it.

*Example:*

```bash
lando start -- -vvvv
info: Bootstraping...
verbose: Emitting event pre-bootstrap
verbose: Loading event post-bootstrap priority 1
verbose: Loading event post-bootstrap priority 2
verbose: Loading event post-bootstrap priority 5
verbose: Loading event post-instantiate-app priority 1
verbose: Loading event post-instantiate-app priority 9
verbose: Loading event post-bootstrap priority 5
verbose: Loading event post-bootstrap priority 1
verbose: Loading event post-bootstrap priority 2
verbose: Loading event post-bootstrap priority 2
verbose: Loading event post-instantiate-app priority 1
verbose: Loading event post-bootstrap priority 2
verbose: Loading event post-bootstrap priority 5
verbose: Loading event post-bootstrap priority 9
verbose: Loading event post-bootstrap priority 5
verbose: Loading event pre-engine-start priority 1
verbose: Loading event pre-engine-start priority 2
verbose: Loading event pre-engine-start priority 2
verbose: Loading event post-instantiate-app priority 1
verbose: Loading event post-instantiate-app priority 5
verbose: Loading event post-bootstrap priority 1
verbose: Loading event poweroff priority 5
verbose: Loading event post-instantiate-app priority 5
verbose: Loading event post-bootstrap priority 2
verbose: Loading event post-bootstrap priority 5
verbose: Loading event post-bootstrap priority 8
verbose: Loading event post-instantiate-app priority 2
verbose: Loading event post-bootstrap priority 1
verbose: Loading event post-bootstrap priority 2
verbose: Loading event post-bootstrap priority 5
verbose: Loading event post-bootstrap priority 9
[...]

PR #1252, opened 2018-10-21
Mount MongoDB data directory as a volume

hussainweb:
The MongoDB service plugin does not add a volume for the data directory for MongoDB. This means that the data does not persist once we run `lando rebuild`.

The changes in this PR also fix the issue in #1086. I have described that particular issue in a [comment on that issue](https://github.com/lando/lando/issues/1086#issuecomment-431690021). With this change, the config option is now similar to other services and should fix the problem.

- [ ] My code includes an update to the [CHANGELOG](https://github.com/kalabox/lando/tree/master/docs/changelog)

If you are unclear about any of the above please add comments below so we can improve our process.

### DIFF
--- a/examples/mongo/.lando.yml
+++ b/examples/mongo/.lando.yml
@@ -26,7 +26,8 @@ services:
 
     # Optionally use the following mongo config file
     # This is relative to the app root
-    config: config.yml
+    config:
+      conf: config.yml
 
   # Spin up services to run a basic node server
   appserver:
--- a/plugins/lando-services/services/mongo/mongo.js
+++ b/plugins/lando-services/services/mongo/mongo.js
@@ -30,12 +30,19 @@ module.exports = lando => {
     // Start a services collector
     const services = {};
 
+    // Define config mappings
+    const configFiles = {
+      conf: '/config.yml',
+      dataDir: '/data/db',
+    };
+
     // Default memcached service
     const mongo = {
       image: 'mongo:' + config.version,
       environment: {
         TERM: 'xterm',
       },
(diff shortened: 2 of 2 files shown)


pirog:
yah this looks reasonable to me as well but like @dustinleblanc my mongo knowledge is not super great. 

This PR is also a great example of how our functional testing suite can really shine in verifying things DWTAFSTD. 

cc @hussainweb @serundeputy

hussainweb:
I wish I could help with tests but my knowledge of Lando isn't much. It took me a long time to even figure this out and this was mainly copied from other db services.

Essentially, MongoDB, like other databases need a place to store the data files. When installed normally, the location is usually `/var/lib/mongo`. For the Docker image, the location is `/data/db`. I got this working at the location, except, like I said, the data didn't persist across `lando rebuild`.

The only issue I see with the change is the schema change for the `config` option. But I think it's better that it matches the schema for other database services rather than keeping it separate here. Also, like I mentioned in #1086, this is not really working anyway.

I'm very happy to help get this moving. Just tell me how.

pirog:
@hussainweb definitely on board with all that you've lined out above, now it's just a matter of getting that test in there so we can verify things. If you want to give it a shot we've got documentation on how to do it (it's actually pretty easy)
https://docs.devwithlando.io/dev/testing.html#functional-tests

And i'm guessing you could probably just augment the current mongodb suite with a test for the volume persistence and custom config file
https://github.com/lando/lando/tree/master/examples/mongo#validation-commands

pirog:
We've got a func test for this now!
https://github.com/lando/lando/tree/master/examples/mongo

PR #1269, opened 2018-10-29
Provide simpletest environment variables

neclimdul:
Should mostly make simpletest work out of the box.

- [ ] My code includes an update to the [CHANGELOG](https://github.com/kalabox/lando/tree/master/docs/changelog)

### DIFF
--- a/plugins/lando-recipes/recipes/drupal7/drupal7.js
+++ b/plugins/lando-recipes/recipes/drupal7/drupal7.js
@@ -83,6 +83,17 @@ module.exports = lando => {
     return build;
   };
 
+  const env = config => {
+    let site = 'https://' + [_.get(config, 'site', config._app), lando.config.proxyDomain].join('.');
+    const env = {
+      SIMPLETEST_BASE_URL: site,
+      SIMPLETEST_DB: 'mysql://' + config._recipe + ':' + config._recipe + '@database/' + config._recipe,
+    };
+
+    // Return the env
+    return env;
+  };
+
   /*
    * Build out Drupal7
    */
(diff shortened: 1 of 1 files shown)


pirog:
@neclimdul this seems like a good idea to me but im wondering if we need to set this at a lower level like the php service because presumably any php driven app could be using simpletest.

@dustinleblanc is this something you could 

1. look into
2. adapt for the new rc2isms?

neclimdul:
@pirog I'm not sure I understand the question but in this case "SIMPLETEST_" in the environment name is a Drupal legacy and drives behaviors in phpunit tests. So this is truly a "Drupal" specific thing.

neclimdul:
From my skimming a while back, I'm not sure there is much directly salvaging this PR with how large the RC2 changes where to related code. I'll follow up in the issue.

pirog:
@neclimdul circling back around on this, where are the `SIMPLETEST` envvars documented? 

I think we should probably adapt this and set them directly in the php service so they are available for all PHP things

neclimdul:
Unfortunately not, they are for both 7, 8, and 9. their name is an artifact of a long dead testing framework.

I'll see if i can find some real documentation but there is some for 8 and 9 in the phpunit.xml.dist file :
https://git.drupalcode.org/project/drupal/-/blob/8.8.x/core/phpunit.xml.dist

pirog:
Ahh nice, thanks for the context @neclimdul. I think we can probably merge this into the drupaly base recipe then

pirog:
This has been adapted and merged into https://github.com/lando/lando/tree/rrc5a

neclimdul:
Thanks for all the awesome work! Love all these cool changes happening

pirog:
@neclimdul thank you! we are very much looking forward to getting that stable release out!

Issue #747, opened 2018-03-09
Recoverable fatal error after switching to php 7.2


robbdavis:
**My set up**
v3.0.0-beta.35 - Mac OS - Sierra

**My `.lando.yml`**
Pantheon recipe - drupal8 framework

After switching to PHP 7.2 I'm getting the following error:

> PHP Recoverable fatal error:  PHP Startup: Cannot set 'user' save handler by ini_set() or session_module_name() in Unknown on line 0

I've switched back to php 7.1 and the error goes away. Drupal is recommending php7.2 now for version 8.5 so this is probably an issue that will become more common as people try updating.

pirog:
@robbdavis is this a problem with lando iself? seems like this is something triggered by your app's code?

robbdavis:
The app is Drupal 8.5. It runs fine on php7.1 but throws this error with php7.2. I also didn't see any errors when I ran the same codebase on Pantheon using php7.2.  That said, maybe I didn't look hard enough. When I get a chance, I'll do some tests again with php7.2 and re-check the pantheon logs.

dustinleblanc:
I'll test this out @pirog. However, Its not an error I see on Pantheon, so I'd imagine if I can repro, we have something goofy going on here.

Google brought this up in a search https://github.com/php/php-src/blob/master/ext/session/tests/session_set_save_handler_class_014.phpt

dustinleblanc:
Can confirm this happens when altering the php value in pantheon.yml:

![image](https://user-images.githubusercontent.com/3948808/37566371-8a4e7d3e-2a8e-11e8-89fa-afd35e0628df.png)

dustinleblanc:
I can confirm this doesn't happen in Drupal 8 recipe

pirog:
@robbdavis @dustinleblanc i suspect this is because the pantheon php image is still on an older version of php (eg 7.2.1). I'm going to rebase the images and im *guessing* that might eb sufficient to close this out.

robbdavis:
Thanks for hunting this down @dustinleblanc and @pirog!

pirog:
@dustinleblanc if you can confirm a `lando rebuild` updates to the latest `7.2.3` then feel free to close this out

dustinleblanc:
@pirog not working on my latest pull of master. still causing the same error to dump to stderr/stdout but it 7.2.3

```sh
Setting up client key /lando/keys/pantheon/bethanys-art-shop/index.pem
[...]

PR #1295, opened 2018-11-16
#747: Fix session save handler

dustinleblanc:
Big props go to @tinefin for finding the root cause of this bug and pinging me to give me the info!

This was apparently a holdover from D6 days that is not friendly with 7.2

Downside is that this might break some d6 sites who will need to add an ini_set in their settings.php


philltran:
@tinefine that is awesome. Thank you. The constant error on every command was driving my OCD nuts.

PR #1260, opened 2018-10-24
Update Terminus 1 9 0

philltran:
Fixes #1259

PR #1308, opened 2018-11-21
passing test for #1260

wenzowski:
I take it from [travis run of my last PR](https://travis-ci.org/lando/lando/builds/457061446) that only unit tests are being run by CI?

This correctly returns a nonzero exit code when the expected version of `terminus` is missing from the appserver container. The double call of `grep` is due to `terminus` passing ansi escape sequences to colorize its output.

As a followup commit, the relevant changelog entry is already present in `master`

### DIFF
--- a/examples/pantheon/README.md
+++ b/examples/pantheon/README.md
@@ -28,7 +28,7 @@ Run the following commands to validate things are rolling as they should.
 cd web && node ../../../bin/lando.js pull --code=dev --database=dev --files=dev
 
 # Check terminus is installed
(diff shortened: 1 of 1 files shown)


dustinleblanc:
We run the functional tests in Circle CI, it looks like those might only be building on PR's that aren't forks though.

wenzowski:
Thought that might be. Had expected last to fail. I don’t have push on this project. Can you copy this over so it triggers @dustinleblanc? TY :)

dustinleblanc:
@wenzowski cook, I am going to close this PR and open one from a branch created fromt he fork

Issue #1058, opened 2018-06-25
Pantheon Recipe doesn't support utf8mb4


dgading:
Bug Report
==========

**Tell us about your setup**

What is your lando version and operating system? **(note that the older your version the less likely we are to reply)**

*v3.0.0-beta.47 on MacOS Sierra*

**Tell us about your `.lando.yml`**

```yml
name: d7-pantheon
recipe: pantheon
config:
  framework: drupal
  site: <site-name> Took these out for client reasons
  id: <site-id> 
  conf:
    database: config/mysql
```

It is also helpful to include other relevant config files. For example, include `pantheon.yml` for a `pantheon` recipe if applicable. Include `php.ini` or `mysql.cnf` if you are overriding our defualt config.

```
[mysqld]
innodb_large_prefix=true
innodb_file_format=barracuda
innodb_file_per_table=truemax_allowed_packet=16M
```
I added the following according to https://pantheon.io/docs/platform-considerations/#emoji-support to my settings.php along with my default `pantheon` database credentials.
```
$databases['default']['default']['charset'] = 'utf8mb4';
$databases['default']['default']['collation'] = 'utf8mb4_general_ci';
```

**Tell us about the command you were running**

```bash
lando start
lando db-import pantheon_live_database.sql.gz
```
Running the steps above returned the following error.
```bash
Destroying all current tables in pantheon... 
NOTE: See the --no-wipe flag to avoid this step!
Gunzipped file detected!
Importing /app/pantheon_live_database.sql.gz...
ERROR 1071 (42000) at line 77: Specified key was too long; max key length is 767 bytes
```
My follow up steps were:
```bash
lando destroy -y && lando rebuild -y
```

During the browser's install steps, when verify requirements appears, I get the following error:

```
Database 4 byte UTF-8 support | Not supported
4 byte UTF-8 for mysql is activated, but not supported on your system. Please turn this off in settings.php, or ensure that all database-related requirements are met. See the [documentation on adding 4 byte UTF-8 support](https://www.drupal.org/node/2754539) for more information.
```

**Tell us generally about your bug**

*I tried to build a local of my Pantheon live site using the Pantheon recipe, but couldn't import the database. I checked for other issues about this subject and could only find #959 which was using the Drupal 8 recipe.*

Please detail the steps we can take to replicate this bug. Be as detailed as possible.

```bash
1. Clone Pantheon D7 site that includes a Database with Collation of utf8mb4_general_ci
2. lando init
3. choose Pantheon recipe
[...]

PR #1345, opened 2019-01-07
Pantheon recipe my.cnf to match pantheon.io config for utf8mb4 support

nikanderson:
This PR relates to #1058 which was closed without a corresponding code change because the submitter resolved his issue locally to his own satisfaction.

Pantheon's MariaDB is configured to support multibyte utf8mb4 via the configuration `innodb_large_prefix=true`. Databases on Pantheon using utf8mb4 may fail to import during `lando pull` or `lando db-import` as a result of this configuration mismatch.

This PR alters the my.cnf file provided by the pantheon recipe to support utf8mb4 by setting `innodb_large_prefix=true` as above, as well as the related `innodb_file_format=barracuda` as required per https://mariadb.com/kb/en/library/innodb-system-variables/#innodb_large_prefix

Default character set or collation are not altered, so existing working configurations should be unaffected.

When the base image updates to MariaDB >= 10.2.2 (currently 10.0.37) these configurations will be correct by default and the options will be deprecated. Until then, this recipe update will improve compatibility with Pantheon.

Thanks!

### DIFF
--- a/plugins/lando-recipes/recipes/pantheon/mysql/my.cnf
+++ b/plugins/lando-recipes/recipes/pantheon/mysql/my.cnf
@@ -102,6 +102,9 @@ innodb_io_capacity      = 512
 innodb_flush_method     = O_DIRECT
 innodb_thread_concurrency = 8
 innodb_lock_wait_timeout = 120
+innodb_large_prefix=true
+innodb_file_format=barracuda
+
 #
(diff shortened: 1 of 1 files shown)


pirog:
@nikanderson thanks a bunch! seems like positive changes to get us even closer to Pantheon parity

Issue #1383, opened 2019-02-02
Postgres doesn't set user/password


dustinleblanc:
When setting up a Laravel recipe, the creds reported by Lando info (and what we would normally set) are user=laravel, pass=laravel, db_name=laravel, but what actually works is user=postgres, pass="", db_name="laravel"

Should be fixable with some env vars per https://hub.docker.com/r/bitnami/postgresql

sarath49:
Been spending 2 days with this being new to lando and needed postgres sql for drupal8 recipe. Unable to connect to database with username=drupal8 password=drupal8 and dbname=drupal8 with host=database and port=5432.

Daisuke-sama:
I have same issue. 
environment variables aren't working and config replacement of lando provides same results
```
    creds:
      database: crmdb
      user: user
      password: secret
```
or
```
services:
  database:
    overrides:
       environment:
         POSTGRES_USERNAME: user
         POSTGRES_PASSWORD: secret
```

Only database name applies.

PR #1384, opened 2019-02-02
Issue #1383: Set Postgres creds

dustinleblanc:
This PR sets the credentials reported by lando info for all Postgres services to match what we're really setting behind the scenes.


pirog:
@dustinleblanc FWIW loading in all of laravel is a pretty expensive test and i think we should try to only use the service itself if possible. Im guessing you can easily enough replicate what a recipe is doing by spinning up another postgres DB and setting its `user` `pass` 

```yaml
services:
  mimicwhatrecipedoes:
    type: postgres
    creds:
      user: something
      database: trouble
      password: somethingelse
```

dustinleblanc:
Refactored the test to be slimmer @pirog

docs/tutorials/postgres.md
```diff
@@ -108,6 +110,7 @@ lando destroy -y && lando start
 
 Also note that by default all `postgres` services have a passwordless `postgres` user with all permissions. **DO NOT ALTER THE PASSWORD OF THE POSTGRES USER.**
 
+Due to the way
```
pirog: @dustinleblanc dont leave me hanging!

pirog:
@dustinleblanc this is looking good. It *might* be worthwhile to update the default config just to help make the default user/pass EVEN MORE obvious

```yaml
services:
  my-service:
    type: postgres:10
    portforward: false
    creds:
      database: database
      # Note that you cannot change the below but they are shown here for transparency
      # You can read more about why this is below
      user: postgres
      password: empty
    config:
      database: SEE BELOW
```

PR #1388, opened 2019-02-03
Fix typo

gormus:
The label should instead read **new** per the description above it.


pirog:
thanks @gormus. out with the old and in with the new

PR #1403, opened 2019-02-04
Fix "Landofile" docs

ro0NL:
looking at 

https://github.com/lando/lando/blob/27974092c6c34d4505045111d9bc694ee641dee0/lib/cli.js#L151-L152

 i think the docs are wrong, this fixes it :)


pirog:
thanks @ro0NL

PR #1453, opened 2019-02-12
Varnish service workspace_client parameter and Pantheon compat

nikanderson:
We encountered an issue with the Varnish config shipped with the Pantheon recipe in Lando. When requesting a page via the edge URL with a large cookie header (in our case around 2500 bytes), Varnish responded with a 500 error while it could be seen via logs that nginx returned 200 OK.

We determined that that problem was as described here: https://www.getpagespeed.com/server-setup/varnish-4-cookie-handling

In short, the Pantheon vcl does a lot of regex-based cookie mangling and needs more working space. According to Pantheon's documentation, the Pantheon Edge size limit for cookies is 10K ( https://pantheon.io/docs/cookies/#cookie-size-limit ). The Pantheon recipe is thus failing out of the box for cookie sizes in excess of around 1/4 the allowed size, or less.

This PR adds a parameter to the Varnish `chaperone.conf` file as recommended by the first link above, doubling the `workspace_client` Varnish parameter from its default value of 64K to 128K.

This change solved our initial issue. We did not test a 10K cookie header size to determine if the maximum allowed cookie size still fails, however.

Thanks!


pirog:
@nikanderson thanks for this! I dont really see a reason for not including this. 

@dustinleblanc, do you have any thoughts on this? It might be nice to try and enforce that 10K limit in the pantheon recipe but that seems like its maybe worth its own issue?

PR #1465, opened 2019-02-14
Minor documentation fix for upgrades to RC2.

danielnitsche:
Example documentation incorrectly lists `service` instead of `services`.


pirog:
nice! good catch @danielnitsche!

PR #1537, opened 2019-03-15
Add support for .mjs file extension by sending correct MIME type

felixarntz:
This PR ensures `.mjs` files are handled properly on Nginx. This file extension is typically used for JavaScript modules (see https://developers.google.com/web/fundamentals/primers/modules), so such files should send the `application/javascript` MIME type header.


pirog:
@felixarntz seems reasonable but i am not super knowledgeable about this. Can you point me to any relevant documentation about possible implications of setting this?

@dustinleblanc you have any knowledge/thoughts about this?

felixarntz:
> Can you point me to any relevant documentation about possible implications of setting this?

I'm not sure about any documentation, but setting this will ensure such files are served with the correct MIME type. Currently, the browser will display an error, because `.mjs` files are not served as `application/javascript`.

dustinleblanc:
I'd normal refer to https://www.iana.org/assignments/media-types/media-types.xhtml for this type of question but I don't see mjs on there. From what I've been reading, the convention is to use .js outside of a Node context (in the browser). @felixarntz do you know what browser support for this is like?

felixarntz:
@dustinleblanc JavaScript modules in the browser are a quite new technology, but it is fairly well supported already, see https://caniuse.com/#search=modules. As stated in the article https://developers.google.com/web/fundamentals/primers/modules, the usage of `.mjs` as file extension is a best practice (indeed taken over from the server-side Node world), but in order to allow browsers to support that format, the server needs to send the correct MIME type `application/javascript`. Especially as Lando is geared towards developers, I think this should be supported, even if it's currently not (yet) in the IANA list you linked to.

dustinleblanc:
Yeah I use the IANA list as a reference only because Pantheon has always done so (and I used to work there, so this kind of thing I default to that way of doing things/knowledge).

It's probably fine to include it.

felixarntz:
What's the state on this? Just following up since it warns about auto-closing :)

pirog:
@felixarntz i think its something we intend to merge but we've just been distracted with other stuff lately :(

felixarntz:
Bump :)

felixarntz:
Bump.

felixarntz:
Bump.

PR #1609, opened 2019-05-08
Update services.md

RichardDavies:
Fix typo in 4th paragraph.


pirog:
ahhh very nice @RichardDavies, i actually did this one as well over at 
https://github.com/lando/lando/commit/9701c575581a202c0a6301e96f984652b920efb8

but no reason why we cant share the cred for it! definitely feel free to directly PR any doc improvements you see, these fixes are really valuable and also save us a TON of time so we can work on bug fixes and new features

PR #1679, opened 2019-06-29
include postgres client default version test

jcandan:
This PR contains a functional test for the `postgresql-client` version default to match `postgres` server default at `10.x`.

- [x] My code includes a [unit test](https://docs.devwithlando.io/dev/testing.html#unit-tests) if applicable

### DIFF
--- a/examples/lamp/.lando.yml
+++ b/examples/lamp/.lando.yml
@@ -1,0 +1,5 @@
+name: lando-lamp
+recipe: lamp
+config:
+  webroot: .
+
--- a/examples/lamp/README.md
+++ b/examples/lamp/README.md
@@ -17,9 +17,13 @@ lando poweroff
 # Should initialize the latest codeignitor codebase
 rm -rf lamp && mkdir -p lamp && cd lamp
 lando init --source remote --remote-url https://github.com/bcit-ci/CodeIgniter/archive/3.1.10.tar.gz --remote-options="--strip-components 1" --recipe lamp --webroot . --name lando-lamp
+rm -rf lapp && mkdir -p lapp && cd lapp
+lando init --source remote --remote-url https://github.com/bcit-ci/CodeIgniter/archive/3.1.10.tar.gz --remote-options="--strip-components 1" --recipe lamp --webroot . --name lando-lamp --option database=postgres
 
 # Should start up successfully
 cd lamp
+lando start
+cd lapp
 lando start
 ```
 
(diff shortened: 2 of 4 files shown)


pirog:
@jcandan always like more tests but i think this should be in the LAMP or php series of tests?

jcandan:
@pirog, this PR is related to #1680. The goal was not to test that it has the client, but that the version matches the `postgres` service's default.

pirog:
@jcandan i get that. my point remains the same though. 

the php and postgres services are decoupled so they cannot influence the contents of one another without introducing what im guessing is very complex and likely brittle build-step based dependencies. so the implication in that is how do we resolve the compatibility issue you raise without violating a key design principle that we adhere to (and enforce downstream in tests) for very practical reasons: it keeps lando stable and maintainable over the long run.

to provide some more concreteness to the above: if the latest version of the postgres-client is backwards compatible with our other supported postgres version then just updating it when it exists in non-postgres services seems like the easiest thing to do. in that case we'd want the tests to be written against _those_ services without a dependency on postgres itself and just check for the right version. note that the functional test suite is very complex so the "fastest, good enough test" is where we should start. If we don't do that the suite will quickly balloon into a duration that is prohibitively long. that said we do want to test the version mismatch explicitly im guessing we'd benefit from testing that in a LAMPy example that already exists so we dont have to add too much overhead. 

another potential route here would be: why are we using the postgres-client inside a php service? would it make more sense to just directly invoke it from a postgres service which presumably would _never_ have the version mismatch? not sure if that is possible because php has all sorts of weird dependencies for its extensions but that would certainly make addressing this problem wayyy easier.

anyway, i hope that is useful feedback!

jcandan:
It is useful feedback, and was originally. I agreed with the need to address this from something other than `postgres`, and I've been looking over the options, and chewing on how to implement this. The tests are directly related to documentation, for which there is a `lamp` and `lemp`, but no `lapp`. The options I've come up with are to either create a `lapp` folder, or to manage a change to `.lando.yml` and `lando rebuild` in the middle of `lamp` tests. What are your thoughts?

jcandan:
Also, yes, the `postgresql-client` is a needed for the `php` extension `pdo_pgsql`.

pirog:
yah makes sense re: `pdo_pgsql`, bummer, guessing thats why it is the way it is to begin with

@jcandan um my initial though would be having the "lamp" test just lando init another lamp stack that uses postgres that way we would only need to grab a single appserver image to power both tests. 

the semantics around the test name remaining "lamp" might need to be reconsidered but i think most people know that "LAMP" means a specific stack but is simultaneously a broader set of similarish stacks.

jcandan:
i'm good with that. let me give it a shot.

pirog:
@jcandan closing this in favor of https://github.com/lando/lando/pull/1699 which was the commits from here

PR #1694, opened 2019-07-11
Fixed reference to non-existent "`run` section"

BWPanda:
Ping @serundeputy


pirog:
@BWPanda for the record `run` _is_ backwards compatible with `build` but def 👍 for having the most up to date docs

Issue #1088, opened 2018-07-09
Idle CPU too high when using Lando compared to other docker containers


tanc:
Bug Report
==========

**Tell us about your setup**

Lando: v3.0.0-beta.47
OS: MacOS 10.13.5

**Tell us about your `.lando.yml`**

```name: drupal-8-core
recipe: drupal8
config:
  webroot: web
  via: nginx
```

**Tell us generally about your bug**

Not a bug but I noticed Lando idles at a higher CPU than the other Docker based environment I have been testing (Docker for Drupal).

When Docker for Mac is first started it's hyperkit process idles at around 2-3% CPU on my machine.

![activity_hyperkit](https://user-images.githubusercontent.com/76320/42446396-fb846364-836d-11e8-8d2b-1ced5b6ceef7.jpg)

When a Drupal 8 recipe based Lando controlled set of containers is started it idles at 15-20% CPU when using NGINX.

![activity_lando_nginx](https://user-images.githubusercontent.com/76320/42446517-7e88511c-836e-11e8-8370-dc98ecd58420.jpg)

The default Drupal 8 recipe using Apache idles at around 7-8% CPU.

I've ssh'd into the app container but its not obvious which process is using CPU.

In comparison I've spun up a Drupal 8 set of containers using Docker for Drupal and this idles at basically the same amount as not having any containers spun up (around 2-3% CPU).

![activity_d4d](https://user-images.githubusercontent.com/76320/42446368-e4d9f804-836d-11e8-9b23-127ebb6acca4.jpg)

Any idea what is causing this CPU load? It would be nice if there was no (extra) load having a Lando site up so that battery lasts longer on laptops and heat is kept to a minimum.

tanc:
This gif shows the nginx and mariadb containers using CPU when idle, using `docker stats`:

![usage nginx and mariadb](https://user-images.githubusercontent.com/76320/42506304-683f9d90-8439-11e8-985c-ff59db081a6f.gif)

`.lando.yml`:
```
name: mytest
recipe: drupal8
config:
  webroot: gitroot/web
  database: mariadb
  xdebug: true
  via: nginx
```

Etroid:
Noticing the same issue when running a `pantheon` D8 recipe.

sherakama:
I am pretty sure this is a Docker/OSX bug. See this thread: https://github.com/docker/for-mac/issues/1759

[...]

PR #1725, opened 2019-08-01
Fixes #1688: Improved healthcheck for Bitnami MySQL containers.

### LINKED ISSUE #1688: Database operations fail immediately after lando starts
**Tell us about your setup**

master (HEAD) as well as v3.0.0-rc.17 on Ubuntu 19.04

```yml
name: dt524
recipe: drupal8
config:
  webroot: docroot
```

We have a custom node.js script that runs lando start, and immediately after it finishes runs Drush commands (`drush uli`, etc) against the container. Use the Drush alias below and run this command (without having previously started this lando container): `lando start && drush @dt524.local uli`

```
local:
  docker:
    service: appserver
    exec:
      options: --user www-data
  uri: dt524.lndo.site
  root: /app
  paths:
    drush-script: /app/vendor/bin/drush
```

**Tell us about the error you got**

Drush fails with `SQLSTATE[HY000] [2002] Connection refused`.

**Tell us generally about your bug**

This is a closely-related follow-up to https://github.com/lando/lando/issues/677 (the error is basically the same). I think the fix that was implemented after that issue is incomplete.

Looking at the logs from the database service creation, I can see that after the mysql service is first started and ready, it takes 1-2 seconds longer for the `drupal8` user to be created. The fix in #677 doesn't use an actual db user for the healthcheck, it only uses the root user, so it reports the container as ready 1-2 seconds before it actually is.

I think the MySQL health check here should use an actual user instead of root. I've verified that this solves the problem, but I'm not sure how to do it in a more general way (I don't know how to access the db creds for the healthcheck).

https://github.com/lando/lando/blob/97986cb0f5dde2c4dc2705fd977a58c43aabbc5e/plugins/lando-services/services/mysql/builder.js#L19

danepowell:

@pirog this fixes #1688 now that the Bitnami MySQL container writes the `mysql_initialized` flag file during setup. The only problem is that _existing_ Lando users running older versions of the Bitnami MySQL image won't have that, and the healthcheck will fail / time out. Do you have any idea how we can prevent that? Is there some mechanism to force people to update their container versions, or to depend on a specific version?

### DIFF
--- a/plugins/lando-services/services/mysql/builder.js
+++ b/plugins/lando-services/services/mysql/builder.js
@@ -16,7 +16,7 @@ module.exports = {
       password: 'mysql',
       user: 'mysql',
     },
-    healthcheck: 'mysql -uroot --silent --execute "SHOW DATABASES;"',
+    healthcheck: 'bash -c "[ -f /bitnami/mysql/.mysql_initialized ]" && mysql -uroot --silent --execute "SHOW DATABASES;"',
     port: '3306',
     defaultFiles: {
       database: 'my_custom.cnf',


danepowell:
I don't think this should be marked as won't fix, it's still an open bug.

pirog:
Nice @danepowell, thanks for scolding the stalebot

danepowell:
I wonder if there's some way for the healthcheck to parse the Bitnami image version. The new healthcheck is supported for versions _higher_ than 5.7.27-debian-9-r12 and 8.0.17-debian-9-r11. The old healthcheck needs to be used for versions prior to that.

pirog:
@danepowell finally able to investigate this more! i think there have been a bunch of changes to bitnami which might make this easier to deal with. Also might be worth looking at #1088 and perhaps devising a healthcheck implementation in Lando that does not run continuously.

pirog:
@danepowell we ended up reimplementing healthchecks around some of the feedback over in #1088 and thats been merged into master with https://github.com/lando/lando/pull/2136. this means it should be in the next edge release. 

Since the delta here is pretty big i think it makes the most sense to close this for now. My recommendation would be to wait for the new edge release and see if this resolve the bug or not. If it doesn't open up an issue and we can go back to the drawing board.

PR #1747, opened 2019-08-15
cert-validity-fix: Fix cert validity on iOS 13+ and macOS 10.15+

colin-campbell:
Lando generated certs after 1st July 2019 are not accepted at all even after trusting the CA on iOS 13 and macOS 10.15+ (giving ERR_CERT_REVOKED in the browser) 
See: https://support.apple.com/en-us/HT210176

This PR restores the cert-generation into compatibility by:
* Adding serverAuth to the extended key usage
* Limiting validity to 825 days.

### DIFF
--- a/plugins/lando-core/scripts/add-cert.sh
+++ b/plugins/lando-core/scripts/add-cert.sh
@@ -18,6 +18,7 @@ cat > /certs/cert.ext <<EOF
 authorityKeyIdentifier=keyid,issuer
 basicConstraints=CA:FALSE
 keyUsage = digitalSignature, nonRepudiation, keyEncipherment, dataEncipherment
+extendedKeyUsage = serverAuth
 subjectAltName = @alt_names
 [alt_names]
 DNS.1 = *.${LANDO_DOMAIN}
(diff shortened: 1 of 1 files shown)


colin-campbell:
Bumping this, because once people start using Mac OS Catalina, new environments will not work.

tom-storms:
+1 for this issue. 
is there a temporary fix to get this working? if i update ~/.lando/scripts/add-cert.sh or ~/.lando/scripts/setup-ca.sh this looks to be overwritten on a lando rebuild

pirog:
@tom-storms this is already in the latest lando release so the workaround is to just update. you _may_ need to just restart your app to get the new cert generated

josephfusco:
@pirog This is still an issue for me in macOS.

macOS 10.15.2
Chrome 79.0.3945.130

.lando.yml
```yml
name: my-project
recipe: lamp
config:
  php: "7.3"
  webroot: "html/"
  database: mysql:5.7
  xdebug: true
```

I can open a new issue if necessary.

PR #1764, opened 2019-08-29
Allow xdebug to connect back automatically

eelkeblok:
It was a bit of a puzzle, but I needed these changes to have xdebug connect back properly to PHPStorm when starting a debug session from the PHPStorm interface (i.e. by pressing the green "bug" button). Without this, no breakpoints would seem to hit, unless I enabled listening manually by clicking the icon with a bug and a phone on it.

- [ ] My code includes an update to the [CHANGELOG](https://github.com/lando/lando/tree/master/help)

### DIFF
--- a/plugins/lando-recipes/recipes/drupal8/php.ini
+++ b/plugins/lando-recipes/recipes/drupal8/php.ini
@@ -30,6 +30,8 @@ date.timezone = "UTC"
 xdebug.max_nesting_level = 512
 xdebug.show_exception_trace = 0
 xdebug.collect_params = 0
+xdebug.remote_autostart = 0
+xdebug.default_enable = 0
 
 ; Globals
 expose_php = on
--- a/plugins/lando-recipes/recipes/drupal8/php.ini
+++ b/plugins/lando-recipes/recipes/drupal8/php.ini
@@ -31,7 +31,6 @@ xdebug.max_nesting_level = 512
 xdebug.show_exception_trace = 0
 xdebug.collect_params = 0
 xdebug.remote_autostart = 0
-xdebug.default_enable = 0
 
 ; Globals
 expose_php = on


pirog:
@eelkeblok not against this per se but AFAIK the default value of both added config keys is already 0. See https://xdebug.org/docs/all_settings. 

So, i'd like to understand how explicitly setting them to the same thing they are set to by default matters since i'm guessing it shouldn't

eelkeblok:
Hmm. That's interesting. I actually think I boiled my changes down to the bare minimum because at some point I was setting everything and it's grandmother because I couldn't figure it out. I'm guessing that probably something else is overrideing the default defaults :) I'll check again to make sure.

eelkeblok:
Ok, so at least further testing proved this only needed [remote_autostart](https://xdebug.org/docs/all_settings#remote_autostart) to be disabled, not sure how I came to [xdebug.default_enable](https://xdebug.org/docs/all_settings#default_enable), must have taken it for something else.

As to how that setting gets changed from the default xdebug setting, not sure. Here's a phpinfo (taken from Drupal's phpinfo screen) that was taken using the vanilla drupal8 recipe:

Screenshot 2019-09-02 at 09 52 15

Although this PR fixes the issue, it may actually be worth digging deeper to see where it is changed from the default in the first place (and why).

eelkeblok:
Apparently they are the standard PHP images from Docker. Not sure why the value gets changes *there*, neither how to find out.

eelkeblok:
@pirog What would be the appropriate way to test for this? Setting first steps in Lando development :)

pirog:
@eelkeblok circling back around to this. whats our status here? what needs to happen to advance this? is this even still applicable?

eelkeblok:
I think we should get a bit more tasks ticked off in the issue summary. I am not sure how best to test for this, though. Do we want to test for the effect (that is going to be very complicated) or just make sure the setting is what we want it to be? Also, I'm not sure if there is any confirmation of the problem. Am I the only one that is an avid user of the "Just click the bug in PHPStorm to start debugging"? Or is something else going on?

PR #1773, opened 2019-09-04
thinktandem/tandem#107: Add installation vid back into the docs.

serundeputy:
@pirog we need this one to get the vid styles; then all the following add vids PRs can just drop the div with the embed code into the `*.md` files.


pirog:
@serundeputy nice. 

we've got responsive youtube video styling on the tandem docs site that might be worth implementing before merging this in.
https://docs.thinktandem.io/

Relevant css is over here
https://github.com/thinktandem/tandem/blob/master/docs/.vuepress/style.styl#L4

Issue #1945, opened 2020-01-21
Using lando generate:guide w/ Team Lando as the author I get an error


serundeputy:
1.  Update to the [latest version](https://github.com/lando/lando/releases) of Lando and make sure the issue isn't fixed already.
 
  * on `master` 

**Tell us about your setup**

  * running lando from `master` with `lando --secret-toggle` on

**Tell us about the command you were running**

When running `lando generate:guide` w/ Team Lando as the author I get an error:

```bash
└─ $ ∴ lando.dev guide:generate
? Guide title? Gff
? Guide description? what gff wants?
? Written by? Team Lando
? Best category? General
? URL of original content? 
? URL of example repo? example_replace
error: pic is not defined
```
```

**Tell us about the error you got**

here is `lando.dev guide:generate -vvv`:

```bash
debug: Emitting event cli-guide:generate-run
debug: Event cli-guide:generate-run has 0 listeners
error: ReferenceError: pic is not defined
    at eval (lodash.templateSources[0]:23:11)
    at Object.run (/home/gff/code/lando-ops/lando/plugins/lando-alliance/tasks/guide-generate.js:121:39)
    at lando.events.emit.then.then.then.then.then (/home/gff/code/lando-ops/lando/lib/cli.js:302:54)
From previous event:
    at lando.bootstrap.then.lando (/home/gff/code/lando-ops/lando/lib/cli.js:300:10)
From previous event:
    at checkForUpdates.then (/home/gff/code/lando-ops/lando/lib/cli.js:268:76)
    at runCallback (timers.js:705:18)
    at tryOnImmediate (timers.js:676:5)
    at processImmediate (timers.js:658:5)
From previous event:
    at Object.handler (/home/gff/code/lando-ops/lando/lib/cli.js:268:42)
    at Object.runCommand (/home/gff/code/lando-ops/lando/node_modules/yargs/lib/command.js:238:44)
    at Object.parseArgs [as _parseArgs] (/home/gff/code/lando-ops/lando/node_modules/yargs/yargs.js:1063:30)
    at Function.get [as argv] (/home/gff/code/lando-ops/lando/node_modules/yargs/yargs.js:1004:21)
    at Cli.init (/home/gff/code/lando-ops/lando/lib/cli.js:211:51)
    at Cli.run (/home/gff/code/lando-ops/lando/lib/cli.js:345:10)
    at Object.<anonymous> (/home/gff/code/lando-ops/lando/bin/lando.js:53:7)
[...]

PR #1946, opened 2020-01-21
#1945: Default guide attributions to Team Lando.

serundeputy:


plugins/lando-alliance/tasks/guide-generate.js
```diff
@@ -114,8 +114,11 @@ module.exports = lando => {
         data.author = author.name;
         data.pic = author.pic;
         data.link = author.twitter ? `https://twitter.com/${author.twitter}` : `https://github.com/${author.github}`;
+      } else if (data.author === 'none') {
```
serundeputy: move this to where the rest of the defaults are set.

Issue #1955, opened 2020-01-28
Cassandra Service?


SpaceBoiNate:
Currently Lando does seem to support MongoDB, however, it would be nice to have Cassandra support as a service. I'm willing to give a stab at creating a pull request, however, I want to make sure that it is worth developing and if there is anything that I should note when developing/adding a service. I've already reviewed the service plugin.

SpaceBoiNate:
Created #1959

serundeputy:
@SpaceBoiNate would you be willing to attend a [`Lando Contributors Meeting`](https://docs.google.com/document/d/1t2lAP3wLiBpaWR90GE7sMT-d6UJ9-GiTe2y50kDuumc/edit#) as the Cassandra representative to give the current state of the PR and determine how to move forward with it?

SpaceBoiNate:
@serundeputy Sorry for the delay, I've been really busy. However, I can look into joining a meeting.

PR #1959, opened 2020-01-30
#1955: Added cassandra service

SpaceBoiNate:
Hello,

This is a quick commit to add the cassandra service to the lando project. I would like to see if there is any feedback from the Lando team on what versions we might want to support, etc. There are still some documentation and ENV vars that need to be exposed and set. However, I think this is a good starting point, as the service loads and works. I'd be happy to work on this so long as I get on some approval to continue.

### DIFF
--- a/plugins/lando-services/services/cassandra/builder.js
+++ b/plugins/lando-services/services/cassandra/builder.js
@@ -1,0 +1,29 @@
+'use strict';
+
+// Modules
+const _ = require('lodash');
+
+// Builder
+module.exports = {
+  name: 'cassandra',
+  config: {
+    version: '3.11.5',
+    supported: ['3.11.5'],
+  },
+  parent: '_service',
+  builder: (parent, config) => class LandoCassandra extends parent {
+    constructor(id, options = {}) {
+      options = _.merge({}, config, options);
+      const cassandra = {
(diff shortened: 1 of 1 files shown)


pirog:
@SpaceBoiNate are you still interested in working on this?

SpaceBoiNate:
@pirog Sure! I know that I need to probably update the documentation and add in some more settings. However, do you know if there is anything else?

pirog:
@SpaceBoiNate, So beyond the builder the two big things are docs and an example we can use for functional testing purposes. 

### Docs

1. Create the main docs page [here](https://github.com/lando/lando/tree/master/docs/config)
2. Update the [services page](https://github.com/lando/lando/blob/master/docs/config/services.md#supported-services) to include a link to the doc above 
3. Update the [menu structure](https://github.com/lando/lando/blob/master/docs/.vuepress/config.js#L161) with the new service

If you want to spin up the docs locally you can follow this
https://docs.lando.dev/contrib/contrib-docs.html

### Example

Add a directory for cassandra to the examples and try to follow the pattern in the other folders for similarish services
https://github.com/lando/lando/tree/master/examples

dont worry about getting it exactly right, as long as someone on the Lando side can generally follow how to set up the service, configure it and test it we can make sure its added to the test suite

Let me know if you need any more guidance!

PR #1995, opened 2020-02-24
Adding Dustins BadCamp 2019 event.

serundeputy:


serundeputy:
@pirog I've added Dustin's BADCamp event in this PR, but I don't see the event in the previous tab on local or P.sh PR env: http://events---pr-1995-ky5a7ay-iaimrn624qfk6.us.platform.sh/

PR #1997, opened 2020-02-24
closes #988 - increase maximum number of returned sites to 5000

### LINKED ISSUE #988: Can't lando init certain Pantheon sites
Bug Report
==========

**Tell us about your setup**

v3.0.0-beta.45 on Mac OSX 10.11.6

n/a (Bug happens when using `lando init` to create a new .lando.yml)

```bash
lando init --recipe pantheon --pantheon-site ls
```

**Tell us about the error you got**

```bash
$ lando init --recipe pantheon --pantheon-site ls --vvvv
? Choose a Pantheon account cboyden@berkeley.edu
error: No such pantheon site!
$ terminus auth:whoami
cboyden@berkeley.edu
$ terminus site:info ls --fields=name,frozen
 ------------ --------------------------------------
  Name         ls
  Is Frozen?   false
 ------------ --------------------------------------
```

**Tell us generally about your bug**
As a Pantheon org admin, I have access to over 1600 sites. When I run `lando init --recipe pantheon` without specifying a site name, only about 270 of them show up in the select list. When I try to pass in a specific Pantheon sitename, any site that hasn't shown up on the select list doesn't seem to be available. This happens even on sites where I am the user in charge.

For the sites that do show up on the select list, the `lando init` command runs fine.

kevcooper:
This HTTP GET parameter (limit) allows Pantheon's API to return more than the default number of results. I'm not actually sure if there is a hard upper bound but I definitely see a difference between the default and adding a limit of 5000.

This should resolve #988

### DIFF
--- a/plugins/lando-pantheon/lib/client.js
+++ b/plugins/lando-pantheon/lib/client.js
@@ -82,7 +82,7 @@ module.exports = class PantheonApiClient {
   getSites() {
     // Call to get user sites
     const pantheonUserSites = () => {
-      const getSites = ['users', _.get(this.session, 'user_id'), 'memberships', 'sites'];
+      const getSites = ['users', _.get(this.session, 'user_id'), 'memberships', 'sites?limit=5000'];
       return pantheonRequest(this.request, this.log, 'get', getSites)
       .then(sites => _.map(sites, (site, id) => _.merge(site, site.site)));
     };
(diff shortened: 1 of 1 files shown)


pirog:
nice @kevcooper! was not aware of this query param! Merged this into a consolidation branch im working on over here: https://github.com/lando/lando/pull/2123 so going to close this in favor of that

thanks a bunch!

PR #2078, opened 2020-03-22
Adds download link for stable Debian release

gitressa:

### DIFF
--- a/docs/basics/installation.md
+++ b/docs/basics/installation.md
@@ -98,8 +98,7 @@ If your Linux distro does not support our `.deb`, `.pacman` or `.rpm` packages y
 
 Note that you *may* also be able to just double click on the package and install via your distributions "Software Center" or equivalent.
 
-Make sure you have wget installed, or install it using:
-`sudo apt install wget`
+Make sure you have wget installed, or install it: `sudo apt install wget`
 
 ### Debian
 
--- a/docs/basics/installation.md
+++ b/docs/basics/installation.md
@@ -98,6 +98,9 @@ If your Linux distro does not support our `.deb`, `.pacman` or `.rpm` packages y
 
 Note that you *may* also be able to just double click on the package and install via your distributions "Software Center" or equivalent.
 
+Make sure you have wget installed, or install it using:
+`sudo apt install wget`
+
 ### Debian
 
 ```bash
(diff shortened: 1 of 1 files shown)


pirog:
@gitressa how do you feel about increasing the scope of this change? I'd love to reorg the linux install docs into a few sections

1. Prereqs (keep the docker community engine install things here)
2. Installation
  a. Download and Click (basically to go github, get the release, double-click on it)
  b. CLI (have one liners people can run to download the latest and install them, would be great to have these for each distro)
  c. eventually a PPAish section?

We can keep the caveats section the same. How do you feel about this?

gitressa:
I am all for that, that sounds like a really great idea! Having something like this would be pretty cool:

`curl -L https://files.devwithlando.io/install_lando.sh | bash`

For inspiration, see https://ddev.readthedocs.io/en/latest/#installationupgrade-script-linux-and-macos

A PPA solution would also be fantastic, and my preferred solution. I understand it will take extra effort from you guys to make it happen, but the benefits for the Lando community will be huge.

pirog:
@gitressa a complete install script might be overkill since presumably this is a two command process eg i think we can continue with what youve suggested in this PR `wget lando && install pkg` but we just want to expand it to all the distros.

We also have hyperdrive which is a scripted way to install from source similar to what you are mentioning. 
https://github.com/lando/hyperdrive

Lets make these proposed doc changes and then maybe spin up another ticket for the PPA?

gitressa:
@pirog `wget lando && install pkg` is an absolutely acceptable solution, let's keep it simple! And opening another issue for PPA makes total sense.

pirog:
@gitressa 100% agreed especially because the `.deb` and other packages do most of the things in the DDEV script already eg install docker/docker-compose, etc. 

only caveat i would add is whether we can assume `wget` is installed on linux by default? i assume so but maybe not?

gitressa:
@pirog I also assumed `wget` was included by default, for example in Ubuntu, but it actually isn't. So we probably need to preface it with:

> Make sure you have `wget` installed, or install it using:
> `sudo apt install wget`

`wget` doesn't require parameters, as opposed to `curl -O`.

gitressa:
I added `wget` instructions and download commands for Fedora and Arch to the PR.

pirog:
@gitressa merge this into a consolidated branch over https://github.com/lando/lando/pull/2114 so closing this. thanks!

gitressa:
Nice @pirog, I am looking forward to the next release, thanks!

PR #2087, opened 2020-03-24
Make steps clearer.

zakiya:
Step two gets lost under the Linux heading. Added more headings to clarify.


pirog:
@zakiya great suggestion! much more readable this way

Issue #1987, opened 2020-02-19
Pantheon Drupal8 missing fastcgi_param PATH_INFO


woodseowl:
Using Lando v3.0.0-rc.23 with Pantheon Drupal 8... 

I'm finding that `PATH_INFO` isn't defined in the server environment variable but there is a line in the pantheon/drupal8.conf.tpl file that suggests it is expected to be:
https://github.com/lando/lando/blob/6a37c0912104861684be3346fa60de76abf38cbc/plugins/lando-pantheon/recipes/pantheon/drupal8.conf.tpl#L154

Based on nginx's documentation, the `PATH_INFO` param needs to be explicitly set:
http://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_split_path_info

This is coming to the surface because [simplesamlphp 1.18 made a change in how it gets `PATH_INFO`](https://github.com/simplesamlphp/simplesamlphp/commit/29b63583d527f568fa8ab2c2d25e11c83f89c33d#diff-af02f7f3daea842b5289008d5d2704c1R136) and the latest drupal/simplesamlphp_auth (v3.2) has recently incorporated that.

I tested locally by running Lando from source and adding the recommended line from the nginx documentation after each of the three `fastcgi_split_path_info` settings:
```
        fastcgi_param PATH_INFO $fastcgi_path_info;
```

This fixed the missing `PATH_INFO` server env variable.

I tried a couple of ways just to set this value through local custom config files (a la https://docs.lando.dev/config/nginx.html#configuration or https://docs.lando.dev/config/pantheon.html#configuration) but couldn't seem to get it to take in this Pantheon D8 setup. It was too much black box experimentation at that point to really try to find it, but I would love to know how to make that work.

dustinleblanc:
Interestingly, I just reviewed the nginx config on one of my Pantheon Drupal sites by doing this:

1. `  lando terminus connection:info localdevux.dev --field=sftp_command`
2. use the provided sftp command to sftp to pantheon
3. run `get nginx.conf` and then `exit`
4. Run `vim nginx.conf` to look at the nginx config file from Pantheon.
5. do a search for `PATH_INFO` in the file

It looks like Pantheon's path info handling more or less matches ours, I grepped for `fastcgi_param PATH` in the file and didn't find anything, so I am wondering if Pantheon set's this value a different way (in the prepend script maybe?)

woodseowl:
[...]

PR #2105, opened 2020-03-27
Issue #1987: Set PATH_INFO and PATH_TRANSLATED for Pantheon Nginx containers

dustinleblanc:
This is done on Pantheon and seems to be needed for the latest versions
of SimpleSaml. Researching these params for NGINX in general, it seems
these values are actually set by default:

@see https://www.nginx.com/resources/wiki/start/topics/examples/phpfcgi/

I've tried to write some func tests for this based on the example above, but I can't get it to work. @pirog I am not sure if the place I editted this is getting loaded correctly.

### DIFF
--- a/docs/help/2020-changelog.md
+++ b/docs/help/2020-changelog.md
@@ -4,6 +4,7 @@
 
 Lando is **free** and **open source** software that relies on contributions from developers like you! If you like Lando then help us spend more time making, updating and supporting it by [contributing](https://github.com/sponsors/lando).
 
+* Set `PATH_INFO` and `PATH_TRANSLATION` fastcgi params for Pantheon [#2105](https://github.com/lando/lando/pull/2105)
 * Added support for `go` versions `1.12-1.14`,
 * Added support for `solr` `8` [#1765](https://github.com/lando/lando/pull/1765)
 * Added new `solr` `7` minor versions [#1765](https://github.com/lando/lando/pull/1765)
--- a/examples/pantheon-drupal7/README.md
+++ b/examples/pantheon-drupal7/README.md
@@ -27,7 +27,6 @@ lando start
 # Should pull down database and files for our drupal7 site
 cd drupal7
 lando pull --code none --database dev --files dev --rsync
-
 ```
 
 Verification commands
--- a/examples/pantheon-drupal7/README.md
+++ b/examples/pantheon-drupal7/README.md
@@ -87,7 +87,7 @@ lando ssh -c "env" | grep TERMINUS_USER | grep landobot@devwithlando.io
 # Should load proper nginx fastcgi params
 cp test.php drupal7
 cd drupal7
-lando ssh -s appserver -c "curl http://appserver_nginx/test.php" | grep "PATH_INFO"
+curl http://landobot-drupal7.lndo.site/test.php | grep "PATH_INFO"
 rm test.php
 
 # Should not set any 8983 perms
--- a/examples/pantheon-drupal7/README.md
+++ b/examples/pantheon-drupal7/README.md
@@ -27,6 +27,7 @@ lando start
 # Should pull down database and files for our drupal7 site
 cd drupal7
 lando pull --code none --database dev --files dev --rsync
+
 ```
 
 Verification commands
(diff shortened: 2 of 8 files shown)


pirog:
@dustinleblanc you got anything i can look at re: tests? Might help me figure out what exactly you are trying to do!

dustinleblanc:
@pirog just put in a WIP test to try to replicate what's listed in the nginx docs

pirog:
Nice thanks @dustinleblanc. So my feedback would be 

* Is it safe to set `PATH_INFO` at such a low level as the `nginx` service? Is this something that is going to behave across ANYTHING we implement using that service? Looks like our Drupal 8 recipe sets it but our WordPress recipe does not. Maybe it's good to set at such a low level but the issue is definitely scoped for _just_ Pantheon i think. 
* It does look like our default nginx pantehon configs have a comment for presumably this exact thing
https://github.com/lando/lando/blob/master/plugins/lando-pantheon/recipes/pantheon/drupal8.conf.tpl#L154 however it also seems like this comment is quickly revealed to be a lie. These configs are adapted directly from Pantheon so you could `sftp` to a pantheon instance and find the updated nginx config
* For the test it might be simpler and more reliable to use `php -i` and just grep for what you need

Hope that helps!

dustinleblanc:
Thanks @pirog! pulling Pantheon config files via SFTP is actually how I landed on this fix. Pantheon has a fastcgi_params file that is nearly identical to ours but includes a couple of Pantheon specific things at the end.

The concern about the global changes to nginx are ones I share. What I think might be better here is creating a new fastcgi_params file in the pantheon recipe, just wasn't sure how to make sure that is loaded correctly. I'll give the php -i route a try and see if I can scope the fastcgi_params

dustinleblanc:
Alright, scoped how we include the prepend.php with a volume and just copied that the for the fastcgi_params file, and then just updated the include path, which is what Pantheon was doing too. Testing it out

pirog:
@dustinleblanc also looking at this more im wondering why we excluded `PATH_INFO` and `PATH_TRANSLATED` from our default nginx `fastcgi_params` file to begin with. Let me git history that file and see if it reveals anything

dustinleblanc:
Okay got the tests passing for this and scoped it to Pantheon in general. I mimic'd their `srv/includes/fastcgi_params` setup on the platform. I can confirm PATH_INFO gets set in the $_SERVER array, but it seems to be empty for the requests I've tested.

Also had to do a slight refactor of the constructor to make code climate happy, that could probably be carried a little further with those functions pulled into the utils in the lib folder.

pirog:
nice @dustinleblanc! super helpful and much appreciated. 

i think we can def just :shipit: this as is. that said and having done some more research i also feel like we might want to revisit our default `nginx` config to include these params as well aka this is the opposite of my original thinking. what do you think? if we feel like that is a good idea i can probably adapt your work here for teh more general case and we can leave pantheon as a special case.

dustinleblanc:
Yeah given that the documentation for nginx _seems_ to indicate that this is just casual normal settings, it feels safe to set everywhere, but I do have that slightly nagging feeling that you were right and something in a far off land, hidden from our view may quietly break 🙈

pirog:
@dustinleblanc ended up just setting the nginx defaults, but merged your other changes in as well. all over in #2123 now! going to close this...

also, no matter what i do i cant seem to get `PATH_INFO` to have a value of any kind... that doesnt seem right to me?

dustinleblanc:
In regards to getting `PATH_INFO` to have actually things, the test scenario from nginx suggests you need something like:

```
// test.php
<?php var_export($_SERVER); ?>

request to -> http://server/test.php/foo/bar.php
```
Then path info should say something like `foo/bar.php` or something.

With our nginx rules that smartly rewrite these requests, I think it just isn't possible to make it show up properly. For most Drupaly recipes I think it just takes that request and routes it to index.php for Drupal to handle.

pirog:
@dustinleblanc i tried doing something similar to that in the `php` example but still got nothing, ditto for drupals `/admin/status/report/php` phpinfo() path. 🤷‍♂ for sure but at least it exists now and we can take it from there

PR #2108, opened 2020-03-28
GUIDES: Addin How do I set the timezone in a container? lando guide.

serundeputy:


pirog:
@serundeputy here are a few suggestions

1. For the title consider "Lando service" instead of "container", this probably increases the SEO factor
2. Instead of doing a more complex bait and switch and `lando rebuild` on the `lemp` example just use one of the more generic service examples and augment eg consider just adding additional `run_as_root` lines to something like https://github.com/lando/lando/tree/master/examples/services

serundeputy:
@pirog I've erradicated the container references in favor of Lando service.

I was unable to get the test to work in the `services-example.func.js` so I put it in the `php-example.func.js`

Added a vid.

smutlord:
consolidated into https://github.com/lando/lando/pull/2112

PR #2118, opened 2020-03-31
Adding an h3 for Good First Issues on contrib/contrib-intro.

serundeputy:


pirog:
good call on using the H2 instead of the H3!

PR #2133, opened 2020-04-06
Add support for Postgres 12.

hailwood:
Hey @pirog as promised here's the PR for postgres 12 support.

Everything seems to be running fine so it was a pretty simple update.
The only thing I'm not sure about is the pinPairs update, I couldn't work out what they're used for.
But I made an assumption and grabbed the latest tag for postgres 12 and added it as a pinPair.

I haven't updated the changelog like the PR requested as well.. I have no idea when this would get released!

### DIFF
--- a/docs/config/postgres.md
+++ b/docs/config/postgres.md
@@ -12,6 +12,7 @@ You can easily add it to your Lando app by adding an entry to the [services](./.
 
 ## Supported versions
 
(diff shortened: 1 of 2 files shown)


pirog:
solid @hailwood, merged this in over here https://github.com/lando/lando/pull/2138

PR #2157, opened 2020-04-14
[guide] How to tokenize all your Pantheon projects

uberhacker:

### DIFF
--- a/docs/.vuepress/guides.json
+++ b/docs/.vuepress/guides.json
@@ -15,14 +15,6 @@
     ]
   },
   {
-    "title": "Node",
-    "collapsable": false,
-    "children": [
-      "frontend",
-      "how-to-wire-up-browsersync-and-gulp-in-my-lando-app"
-    ]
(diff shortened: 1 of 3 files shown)


pirog:
@uberhacker AFAIK the assumption of this article that you cannot use the same token for more than one project is not accurate. The obvious counterexample is Pantheon Localdev itself which requires the user put in one token and then allows the user to pull any site from the account associated with that token.

uberhacker:
It is true that you can use the same token for more than one project.  However, it's not true that you can use the same token for multiple Pantheon accounts (ie. login with a different email address).  That is the point I was trying to make clear.  Maybe it needs to be reworded?

uberhacker:
Pantheon tokens are stored in `~/.lando/cache/pantheon.tokens` so this isn't really needed.  That said, if this file is removed, Lando will grab the tokens in `~/.terminus/cache/tokens` on your local machine and use them in the containers.

PR #2162, opened 2020-04-16
Only add PathPrefixStrip if we actually have a subdir.

### LINKED ISSUE #2149: compose service proxy gives ERR_INVALID_REDIRECT
`v3.0.0-rrc.3`, `v3.0.0-rrc.4` (or latest Docker shipped with it) breaks one of my compose services, namely adminer - it's proxy gives ERR_INVALID_REDIRECT after the upgrade. Setup has been working until `v3.0.0-rrc.2` (including).

Reference to @pirog's suggestion on Slack to file an issue: https://devwithlando.slack.com/archives/C2XBSHX8R/p1586678412457800

**Tell us about your setup**

*v3.0.0-rc.4 on macOS Catalina 10.15.4 (19E287)*

```
services:
  adminer:
    type: compose
    services:
      image: dehy/adminer
      command: /bin/s6-svscan /etc/services.d
    portforward: true
proxy:
  adminer:
    - adminer.lndo.site
```

The whole project and `adminer` service setup: https://github.com/wunderio/drupal7-project/blob/master/.lando.yml#L8-L16

Service's proxy is green after the build, but it gives before mentioned `ERR_INVALID_REDIRECT` error when trying use it (to log in to Adminer via UI).

Not sure if this is Lando's bug, but let's start here ;)

JorisVanEijden:
parsedUrl.pathname is '/' if there is no subdir, but that should not be sent as a stripped prefix.

Fixes #2149

### DIFF
--- a/plugins/lando-proxy/lib/utils.js
+++ b/plugins/lando-proxy/lib/utils.js
@@ -69,7 +69,7 @@ exports.parseRoutes = urls => {
     const hostRegex = parsedUrl.host.replace(new RegExp('\\*', 'g'), '{wildcard:[a-z0-9-]+}');
     labels[`traefik.${i}.frontend.rule`] = `HostRegexp:${hostRegex}`;
     labels[`traefik.${i}.port`] = parsedUrl.port;
-    if (parsedUrl.pathname) {
+    if (parsedUrl.pathname.length > 1) {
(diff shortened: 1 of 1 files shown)


pirog:
Nice work @JorisVanEijden!

PR #2236, opened 2020-05-09
cache: none gave me an error but when I removed it worked fine

chapmanjacobd:
before: Cannot read property 'type' of undefined

### DIFF
--- a/docs/config/laravel.md
+++ b/docs/config/laravel.md
@@ -51,7 +51,6 @@ config:
   via: apache:2.4
   webroot: .
   database: mysql:5.7
-  cache: none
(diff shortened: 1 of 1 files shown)


smutlord:
@chapmanjacobd i dont think we want to remove the entry since its important to communicate that the cache is off by default. that said you are right that we should not use the string literal "none"

smutlord:
@chapmanjacobd on second thought i think this should be a bug and `none` should be something lando can handle

smutlord:
@chapmanjacobd this has been merged over here https://github.com/lando/lando/pull/2245/files

Issue #2246, opened 2020-05-12
Add support for MongoDB 4.2


rikvanderkemp:
**Please explain your request and its importance in the form of a user story.**

As a lando user who uses MongoDB, I would like to have the latest 4.2 available as a service. 4.1 seems to have no official stable release and 4.0 is a old release.

**Steps**

- [x] add the latest [bitnami mongodb](https://github.com/bitnami/bitnami-docker-mongodb/tree/4.2.6-debian-10-r33/4.2) container to the service provider
- [x] update documentation https://docs.lando.dev/config/mongo.html#configuration
- [x] add a unit test for the latest version to verify implementation

**Caveat**

Container is only available with Debian 10. I do not foresee issues here this is in a similar context of the PHP 7.4 container (which is also using Debian 10)

pirog:
@rikvanderkemp have you tried overriding Lando and using the 4.2 image to see what happens?

rikvanderkemp:
I just finished on code changes to make it work. So I'll be submitting a PR in a minute if that is alright?

pirog:
@rikvanderkemp perfect

rikvanderkemp:
Can I add you as the pinged project committer @pirog ?

smutlord:
@rikvanderkemp im on it. will post some comments on the PR

PR #2247, opened 2020-05-12
lando#2246: add new container reference, update docs and add unit test

rikvanderkemp:
- [X] My code includes the latest code from the `master` branch
- [X] My code includes a [unit test](https://docs.lando.dev/contrib/contrib-testing.html) if applicable
- [X] My code includes documentation updates if applicable.
- [ ] I have pinged [a project committer]


plugins/lando-services/services/mongo/builder.js
```diff
@@ -8,8 +8,9 @@ module.exports = {
   name: 'mongo',
   config: {
     version: '4.0',
```
smutlord, line 10: @rikvanderkemp based on your comments on #2246 re: status of 4.0 and 4.1 does it make sense to set 4.2 as the default version?

rikvanderkemp: Great point! 

Especially when I look at the following page:
 https://www.mongodb.com/support-policy

4.0 and 4.2 are both TBD while 4.1 isn't mentioned.

4.0 is from June 2018 and 4.2 August 2019.

Based on this I would say set 4.2 to default.

Maybe even dropping 4.1 altogether and add a deprecation warning for it. @pirog did something similar to phpmyadmin.

What do you think?

pirog: @rikvanderkemp was looking at exactly the same. yes i think moving 4.1 into the "you can use it still but it throws a warning" group makes sense.

im merging a bunch of stuff together rn so im happy to make these changes

rikvanderkemp: Perfect that would be great.

smutlord:
@rikvanderkemp ive got this over on this branch now. would love to get your review
https://github.com/lando/lando/pull/2251

Issue #2308, opened 2020-06-01
Spin off overrides.js so we have an easy place to track `PLATFORM_VARIABLES` that need to be modiest


pirog:
There are some framework specific settings (eg the location of the temp directory for Drupal 8) that are set to less than ideal things by default. If any of these settings are dealbreakers for local dev we should override them and document that we do so _for now_.

- [ ] Create `overrides.js`
- [ ] Set D8 things we need for now
- [ ] Spin up ticket with below discussion

Longer term we should figure how we want to handle this problem. Some ideas:

1. Update `platform.sh`s config files eg `settings.platform.php` to set LANDO things as needed
2. Continue to override `PLATFORM_VARIABLES` with what we need
3. Tell the user they need a separate lando config file, provide example or "starter" file
4. Some combination of the above.

PR #2324, opened 2020-06-02
#2308: Spin off, implement for D8 and document PLATFORM_VARIABLES ove…

pirog:
@mikemilano:

* reorganized this a bit better
* set a smaller amount of things for D8 for now
* documented it in the user docs

### DIFF
--- a/docs/config/platformsh.md
+++ b/docs/config/platformsh.md
@@ -311,7 +311,7 @@ Then you'd expect to see the following commands and usage:
 
 ```bash
 lando database  Connects to the database relationship
-lando redis  Connects to the database relationship
+lando redis     Connects to the database relationship
 ```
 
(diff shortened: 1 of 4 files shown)


docs/config/platformsh.md
```diff
@@ -311,7 +311,7 @@ Then you'd expect to see the following commands and usage:
 
 ```bash
 lando database  Connects to the database relationship
-lando redis  Connects to the database relationship
+lando redis     Connects to the database relationship
 ```
```
mikemilano, line 315: This line added spaces I'm not sure were meant to be added.

pirog: @mikemilano yup, intentional, cosmetic change so the command descriptions are in the same column

PR #2335, opened 2020-06-04
CA: Ensure cert and key owned by host user/group

diddledan:
Add chown commands to `setup-ca.sh` and `add-cert.sh` to ensure that the created certs and keys are correctly owned by the host user and group ID. I've tested this on Ubuntu where everything seems to be good, but haven't tried Windows yet to see if there are any regressions.


diddledan:
@pirog, @serundeputy, @dustinleblanc I think this is good to go. I haven't added myself as a contributor though, because I'm unlikely to be able to commit much time.

pirog:
@diddledan we arent sure what this PR is all about. Does this fix a known bug? It _seems_ reasonable but it also is likely to cause regressions elsewhere. Can you give us a deeper explanation around the change?

diddledan:
Hi, sure.

The CA certificates are created owned by `root` because that is the user inside the container that generates the certificates. On Linux this root ownership is propagated to the ownership on the host filesystem. This means that on the host filesystem the CA certificates are owned by root and not the logged-in user's account. This is unexpected but in of itself not a problem in normal use.

However, I've been working towards packaging Lando into a [Snap Package](https://snapcraft.io/) for Linux. Here, the restrictions imposed by the MAC (AppArmor) confinement mechanisms mean that the root-owned CA certificates are prevented from being accessed by Lando running as the host's logged-in user thereby breaking all sorts of things with unhelpful messages due to the unexpected nature of the problem.

This change will run `chown` inside the CA generating container to change the ownership inside the container to that of the host's logged-in user ID. That change of ownership is then propagated to the host filesystem in the same way that the current root ownership is. The change fixes the Snap Package use-case, and I've tested that scenario.

I'm just re-checking now to see if Windows is broken at all due to the different filesystem semantics.

pirog:
@diddledan are you planning on packaging in the root CA or something? I don't understand what the CA has to do with the SNAP package.

diddledan:
The CA is generated by Lando at runtime the first time a user creates a site. The CA is generated per system due to the requirement for having access to the private key to sign the site certificates. This change also affects the certificates for each site that are signed by the CA, which are also owned by root when they are created by Lando.

Also, I have just successfully run the changed code to initialise and start a test site without any issues on Windows - so I didn't break Windows usage :-).

pirog:
@diddledan that doesnt really answer my question. CA generation should happen AFTER lando has been installed, as you point out, so unless you are generating it during the SNAP install process somehow i dont see why this would be problematic.

We aren't against this change but we are trying to understand the cost/benefit here. This is actually a pretty deep change that doesn't fix any currently known issues with Lando so we'd prefer not changing it unless there is a compelling reason to do so.

diddledan:
Snap Packages are executed inside a sandbox to restrict their ability to access the rest of the system they are running upon. Think of this concept like the sandbox that an iOS app is placed within when running on an iPhone. The Access Control software that mediates filesystem access is called AppArmor. When a user installs my experimental Lando snap they do not get a CA certificate, the same as installing Lando from any other source. When they execute Lando via the Snap Package to start a new site Lando will start several containers via Docker to generate the SSL certificates and the CA that signs them. These are then saved by the containers to `$HOME/.lando/certs` (more specifically in the Snap view of the world this is an alternative `$HOME` at `$HOME/snap/lando/current/.lando/certs`). The openssl commands being run by those containers to generate the certificates are executed using the root user with UID 0 inside the container so the certificate files are saved with root as their filesystem-defined owner. These ceritificates are saved inside the container to a path that has been mounted from the host that maps to the `$HOME/.lando` folder. Because of the mounting of the `.lando` from the host the files are saved to the `.lando` folder with the same root/uid-0 filesystem-defined owner as they have inside the containers.

The AppArmor confinement now comes into play because Lando executing on the host outside of the containers is executing as (for example) `diddledan`. Because Lando is running as `diddledan` AppArmor will refuse to allow Lando access to the newly generated certificate files in `$HOME/.lando/certs` because they are owned by a user that is not `diddledan`, i.e `root`. Lando will then receive an EPERM error from the kernel when it accesses the certificate files because AppArmor is denying the access attempt. When Lando fails to read the new certificates, it will throw error messages and the site the user is attempting to launch will not be functional.

The change here alters the script that is executed inside the containers that generate the certificates (both site and CA) to call `chown` to change the ownership of the new files to that of the user who is trying to run Lando. This will appease the AppArmor restrictions because the files will now be owned by `diddledan` (in my example) which is the expected user so the file access is allowed. The site will successfully start and the user can begin their work.

dustinleblanc:
@diddledan You mentioned the snap changing the $HOME variable, and I feel like that is going to cause all sorts of problems with the fundamental assumptions of how Lando operates. We rely on mounting the user's $HOME directory to access things like .ssh keys, etc. How does that work with a Snap package?

diddledan:
According to https://docs.lando.dev/config/ssh.html Lando only manages its own path for SSH Keys at `.lando/keys`. It looks like there is the ability to copy keys from the user's `$HOME/.ssh` directory. These can easily be read-only accessed using the appropriate permissions to access SSH keys from the user's real `$HOME` via a Snap permission.

It is a tenet of the Snap ecosystem that Snaps be granted to minimum access they require and allow the user to opt-into granting further permissions if they need them. IMO, the SSH keys access would fit into this opt-in paradigm with access to `$HOME/.ssh` denied by default - this would not prevent Lando managing existing keys in `.lando/keys`.

However, SSH Keys are unrelated to this PR, and can be investigated once the hard block of being completely unable to start a site when Lando is contained in a Snap package that this PR will fix.

dustinleblanc:
@diddledan I just want to make sure the end goal of Snap distribution support jives with all of Lando's features. If we are going to make changes to allow snap packaging to work, I want to make sure that users who install the snap can get the full raw awesomeness of what Lando provides. I am all for making Lando easier to install, and tighter security is also a valiant goal, but as a project we've often erred on the side of easier to use and more powerful rather than worrying about things like process sandboxing, hence we mount user's SSH keys so that things like `lando drush @server.site do_thing` just work out of the box. We want to make sure that experience doesn't break if we go down the road at all.

@pirog and I are doing a risk/reward calculation in our heads about the overall goal's benefits vs the potential for something to stop working correctly, that's all.

It sounds like in general, allowing access to files in the user's home directory is possible, but the moving of all of Lando's config/cache and other files to somewhere that is not the equivalent of the user's home directory could cause some breakages, and if we were to pursue supporting that paradigm, we might want to find a way to run our functional test suite using that directory structure.

Issue #2318, opened 2020-06-02
Provide better guards in `recreate-users.sh`


pirog:
Right now our `recreate-users` script is "bad" in that it always runs and never errors. We should only attempt to run the commands in this script when we need to and they should actually report failure if they fail in those circumstances. 
https://github.com/lando/lando/blob/master/experimental/plugins/lando-platformsh/scripts/psh-recreate-users.sh 

The guards we need are 

1. If `web` user exists and is set correctly (eg has UID = `$LANDO_HOST_UID` and correct HOME and SHELL) then we dont need to recreate `web`
2. If `app` user exists and is NOT set to `$LANDO_HOST_UID` then we do not need to recreate. 

Note that we need to end up with a result where the `web` user has `$LANDO_HOST_UID` and the `app` user does not. This can be tricky because on `linux` user `1000` is common and `app` is set to `1000` by default. This means we need to remove the `app` user first, then create `web` next so that then we create `app` its not going to be to `1000` if `$LANDO_HOST_UID` is also `1000`

mikemilano:
Updates in PR in #2338

It works successfully with a `lando rebuild` and `lando restart`.

PR #2338, opened 2020-06-04
#2318 Updated script to assure useradd wasn't executed if web user 50…

mikemilano:
This PR contains an update of `psh-recrate-users.sh` which does not execute any add or delete user functionality if the `web` and `app` users are setup appropriately.

In the case the `web` is not uid `501`, or that uid `501` is not `web`, the user will be deleted and recreated appropriately.

### DIFF
--- a/experimental/plugins/lando-platformsh/scripts/psh-recreate-users.sh
+++ b/experimental/plugins/lando-platformsh/scripts/psh-recreate-users.sh
@@ -1,5 +1,5 @@
 # This script assures there is a `web` and `app` user
-# and that the `web` user is assigned the uid of 501.
+# and that the `web` user is assigned the uid of $LANDO_HOST_UID.
 #!/bin/sh
 set -e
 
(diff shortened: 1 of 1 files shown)


experimental/plugins/lando-platformsh/scripts/psh-recreate-users.sh
```diff
@@ -1,3 +1,5 @@
+# This script assures there is a `web` and `app` user
+# and that the `web` user is assigned the uid of 501.
```
pirog, line 2: @mikemilano this all looks great and very much appreciate the comments to help me walk through the logic. Only suggestion is to modify 

> and that the `web` user is assigned the uid of 501.

Since 501 is not universal and probably only true for the first user account on a mac

mikemilano: good catch, this has been updated.

pirog: nice. thank you kind sir!

Issue #2376, opened 2020-06-09
Expose `psysh` for lando-platfromsh php recipes.


serundeputy:
The `lando-platformsh` (at least for php apps) has `psysh` in the app container.  This tooling works to expose it to the user:

```yaml
tooling:
  psysh:
    service: app
```
Can we expose that to the user by default for PHP apps?

PR #2377, opened 2020-06-09
#2376: Adding psysh tooling to platformsh php apps.

serundeputy:
Looks like tests are just a stub of lagoon so far, so i didn't add any tests yet.


pirog:
nice taking the initiative here @serundeputy but im going to hold off on merge before we get feedback from platform. im not sure if we want to include docs or other things.

PR #2438, opened 2020-07-01
Rabbitmq

mikemilano:
- Build fails after last test completes and so I'm unsure why that is
- Kafka hangs when building locally; The process is running however.
  - We reviewed the `my_ip` error in the lando logs and it was deemed that it should not be a problem
  - I commented the Kafka service out in kitchensink & functional tests


examples/platformsh-kitchensink/README.md
```diff
@@ -237,6 +237,25 @@ lando ssh -c "curl localhost/solr.php" | grep "Result" | grep "OK"
 # Should run php mongodb commands successfully
 cd sink/php
 lando ssh -c "curl localhost/mongodb.php" | grep "Result" | grep "OK"
+
+# Should have OS Pid when running rabbitmqctl status
```
pirog, line 241: @mikemilano i thin you probably need to `cd sink/php` here

examples/platformsh-kitchensink/README.md
```diff
@@ -237,6 +237,25 @@ lando ssh -c "curl localhost/solr.php" | grep "Result" | grep "OK"
...
+lando ssh -c "curl localhost/rabbitmq.php" | grep "Result" | grep "OK"
+
+# Should show influxdb process running as the app user
+cd sink/php
+lando ssh -u root -s influxdb -c "ps aux|grep influxdb" | grep "^app"
+
+# Should show kafka process running
+# Temporarily commented out until kafka issues are resolved
+# lando ssh -u root -s kafka -c "ps aux|grep '\srunsv kafka'"
 ```
```
pirog, line 259: @mikemilano you might need to throw in a dummy command here like 

```md
# Should show kafka process running
# Temporarily commented out until kafka issues are resolved
# lando ssh -u root -s kafka -c "ps aux|grep '\srunsv kafka'"
true
```

im not sure if the test framework knows how to parse a block that is all comments correctly

experimental/plugins/lando-platformsh/services/platformsh-varnish/builder.js
```diff
@@ -0,0 +1,38 @@
...
+      const varnish = {
+        image: `docker.registry.platform.sh/varnish-${options.version}`,
+        ports: [options.port],
+        environment: {
+          LANDO_WEBROOT_USER: options.meUser,
+          LANDO_WEBROOT_GROUP: options.meUser,
+        },
+        networks: {
+          default: {
+            aliases: ['varnish'],
```
pirog, line 29: @mikemilano what is the purpose of the alias usage here and in the other services in this PR? All the services should automatically get `NAME.internal` aliases plus the ones lando gives by default.

mikemilano: I had copied the networks block into a few services as I stubbed them out. I removed them all.

PR #2441, opened 2020-07-02
2402 better pantheon pull

dustinleblanc:
This should simplify the database portion of Pantheon pulls and should result in faster pulls for all users, especially in the live environment on Drupal sites. We're relying on Drush to just skip cache data rather than flushing the cache first.

I manually tested this with D7, D8, WP, and WP Networks and it seems to be working fine. I have not tested with Backdrop

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
       image: ubuntu-1604:201903-01
       # @TODO:: once we have more tests running lets buy the below and
       # set up max parallelism
-      # docker_layer_caching: true
+      docker_layer_caching: true
     parallelism: 20
     steps:
       - checkout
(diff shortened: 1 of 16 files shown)


uberhacker:
Nice @dustinleblanc!  Is this a noticeable performance improvement?  I see you are still using my fu. LoL

dustinleblanc:
Yeah @uberhacker depending on the size of the database, the diff can be significant.

pirog:
This is benchmarked against a vanilla D7 database: one of the first two are what lando does currently based on a bunch of things. the third is what this moves to.

```bash 
# mysqldump
time mysqldump --no-autocommit --single-transaction --opt -Q -u pantheon -pMYPASSWORD -h dbserver.dev.SITEID.drush.in -P 17242 pantheon

real	0m53.512s
user	0m0.057s
sys	0m0.201s

# drush sql-dump
time drush @pantheon.landobot-drupal8.dev --extra-dump=--column-statistics=0 sql-dump

real	0m12.732s
user	0m0.570s
sys	0m2.048s

# terminus remote:drush
time terminus remote:drush -- sql-dump --extra=--column-statistics=0

real	0m4.939s
user	0m0.233s
sys	0m0.082s
```

PR #2464, opened 2020-07-13
Add triaging issues documentation

dustinleblanc:


dustinleblanc:
Move Zenhub usage to an optional section (appendix, etc)

dustinleblanc:
Move page to its own menu section "Triaging Issues > Getting Started" or see if we can just make it a top level doc with header navigation

PR #2467, opened 2020-07-14
Added yaml support for the include tag.

mikemilano:
This PR adds support for the !include yaml tag used in the varnish service.


pirog:
merged this into a consolidated branch im working https://github.com/lando/lando/pull/2469

Issue #2428, opened 2020-06-28
Unhandled rejection SyntaxError: Unexpected token  in JSON at position 0


jeffymahoney:
3.0.7 on Mac Mojave

When I try any lando command from any path, including just running `lando` with no arguments, I get the following error:

```
$ lando
Unhandled rejection SyntaxError: Unexpected token  in JSON at position 0
SyntaxError: Unexpected token  in JSON at position 0
    at JSON.parse (<anonymous>)
    at /snapshot/lando/build/cli/plugins/lando-pantheon/lib/utils.js
    at arrayMap (/snapshot/lando/build/cli/node_modules/lodash/lodash.js:639:23)
    at Function.map (/snapshot/lando/build/cli/node_modules/lodash/lodash.js:9556:14)
    at interceptor (/snapshot/lando/build/cli/node_modules/lodash/lodash.js:16993:35)
    at thru (/snapshot/lando/build/cli/node_modules/lodash/lodash.js:8797:14)
    at /snapshot/lando/build/cli/node_modules/lodash/lodash.js:4374:28
    at arrayReduce (/snapshot/lando/build/cli/node_modules/lodash/lodash.js:683:21)
    at baseWrapperValue (/snapshot/lando/build/cli/node_modules/lodash/lodash.js:4373:14)
    at LazyWrapper.lazyValue [as value] (/snapshot/lando/build/cli/node_modules/lodash/lodash.js:1859:16)
    at baseWrapperValue (/snapshot/lando/build/cli/node_modules/lodash/lodash.js:4371:25)
    at LodashWrapper.wrapperValue (/snapshot/lando/build/cli/node_modules/lodash/lodash.js:9052:14)
    at Object.getTerminusTokens (/snapshot/lando/build/cli/plugins/lando-pantheon/lib/utils.js)
    at getTokens (/snapshot/lando/build/cli/plugins/lando-pantheon/recipes/pantheon/init.js)
    at Object.options (/snapshot/lando/build/cli/plugins/lando-pantheon/recipes/pantheon/init.js)
    at /snapshot/lando/build/cli/plugins/lando-recipes/lib/options.js
From previous event:
    at processImmediate (internal/timers.js:439:21)
```

**Tell us more**

I have tried uninstalling and reinstalling Lando (including removing ~/.lando) and Docker multiple times.
I've tried with 3.0.7, 3.0.5, and 3.0.1.
I've searched the issue queue here and haven't found anything in all the issues ever filed.
[...]

PR #2468, opened 2020-07-14
Issue #2428: Catch JSON parse errors in Terminus tokens

dustinleblanc:


pirog:
@dustinleblanc closing this because i merged it in over here https://github.com/lando/lando/compare/308a?expand=1

PR #2522, opened 2020-07-26
Show app name in Destroy prompt

gitressa:
**Problem**
Sometimes you copy `.lando.yml` files from project to project, and forget to update the `name` parameter, which could then overwrite as well as remove the wrong project, when you run `lando destroy`.

**Solution**
It would be great if the app name was shown as part of the Warning message after running `lando destroy`, since this could help catch such instances, before it was run.

This might not work at all, since I can't test it locally ... I tried to find the Lando files with `whereis lando`, which pointed me to `lando: /usr/local/bin/lando /usr/share/lando` (Ubuntu 18.04) but neither place, seemed to contain the `destroy.js` file ... if anyone can point me to the correct folder, I'd be thankful. Possibly `const app = lando.getApp(options._app.root);` needs to be added, for `app.name` to be available?

### DIFF
--- a/plugins/lando-core/tasks/destroy.js
+++ b/plugins/lando-core/tasks/destroy.js
@@ -5,7 +5,7 @@ module.exports = lando => {
     command: 'destroy',
     describe: 'Destroys your app',
     options: {
-      yes: lando.cli.confirm('Are you sure you want to DESTROY?'),
+      yes: lando.cli.confirm('Are you sure you want to DESTROY %s?', app.name),
     },
     run: options => {
       // Stop rebuild if user decides its a nogo


pirog:
@gitressa you need to install lando from source in order to be able to modify the code
https://docs.lando.dev/basics/installation.html#from-source

gitressa:
Thanks @pirog, that was easier than I feared. To answer my question, my PR doesn't work, but I still think it's a great idea. Unfortunately I have no clue how to adjust the actual code ...

pirog:
yup! installing from source isnt that bad. on the PR fron: makes sense, since `app` is not defined until later, you do have access to `lando` though so its possible you might be able to get the app name from something in there

gitressa:
This issue is still relevant.

Issue #2529, opened 2020-07-30
Add --destination support to platform local build when passthru file does not exist


mikemilano:
When using the Drupal 7 p.sh template (and potentially others), `web.locations` (in `.platform.app.yaml`) path to `index.php` does not exist. As a result, Lando cannot serve the site.

`_www` is the default destination created by `platform build` however using this adds other concerns in breaking with the definitions of `web.locations`.

## Solution

- On `lando build`, if `web.locations./.passthru` has a file defined, and that path does not exist, set the path as the `--destionation` value on `platform local:build` in `psh-build.sh` (line 35)
- Add p.sh destination path support to `.lando.yml`
  - This will be an undocumented escape hatch to override anything we've set, or cover any case we haven't

## Notes

- p.sh config data in `integrations/lando-platformsh/app.js` -> `app.platformsh`
- `.platform.app.yaml` is parsed and set in `config.js` -> `parseApps` 
- `lib/run.js` -> `getDocRoot` may work to identify the root path for 90%+ cases

If we go with passing as an argument
- 'lib/services.js` is where `helpers/psh-build.sh` is called
  - add `--destination`

If we go with env vars
- Set it in `types/platformsh-appserver/builder.js`

pirog:
@mikemilano long term we will eventually move to use `/etc/platform/commands/build` but i think that might be a post-beta activity so lets def do this for now.

PR #2530, opened 2020-07-30
#2529: Use destination parameter if needed when doing platform build

pirog:
@mikemilano the check for this ended up being a lot easier than we had originally thought. We didnt need to check for the entrypoint just whether the webroot existed or not.


mikemilano:
@pirog Tested locally on D7 after deleting the existing symlinks. Works as expected.

Issue #2504, opened 2020-07-20
Skip permission hardening in Pantheon recipe automatically


dustinleblanc:
When running locally, a Drupal develop pretty much NEVER wants to run Drupal's permission hardening. It causes headaches when trying to commit files in the sites/default directory that you have changed. We already allow overriding this on the platform recipe, and it looks like we can probably fairly easily support this (at least in D8+) in the Pantheon recipe with something like the following:

```js
const getPantheonSettings = options => ({
  databases: pantheonDatabases,
  conf: {
    'pressflow_smart_start': true,
    'pantheon_binding': 'lando',
    'pantheon_site_uuid': options.id,
    'pantheon_environment': 'lando',
    'pantheon_tier': 'lando',
    'pantheon_index_host': PANTHEON_INDEX_HOST,
    'pantheon_index_port': PANTHEON_INDEX_PORT,
    'redis_client_host': PANTHEON_CACHE_HOST,
    'redis_client_port': PANTHEON_CACHE_PORT,
    'redis_client_password': PANTHEON_CACHE_PASSWORD,
    'file_public_path': 'sites/default/files',
    'file_private_path': 'sites/default/files/private',
    'file_directory_path': 'sites/default/files',
    'file_temporary_path': '/tmp',
    'file_directory_temp': '/tmp',
    'css_gzip_compression': false,
    'js_gzip_compression': false,
    'page_compression': false,
  },
  drupal_hash_salt: getHash(JSON.stringify(pantheonDatabases)),
  config_directory_name: 'config',
  settings: {
    skip_permissions_hardening: 1,
  },
});
```

(source is here: https://github.com/lando/lando/blob/master/integrations/lando-pantheon/lib/utils.js#L85)

We should investigate adding the skip_permissions_hardening here to see if we can alleviate some problems users of both Lando and Localdev will have with modifying files like settings.php

Pinging @pirog for the visibilities

pirog:
@dustinleblanc here is where the magic happens on D8
https://github.com/pantheon-systems/drops-8/blob/28990c96f2bc611d95bf87696b8c51bfa096f6c4/sites/default/settings.pantheon.php#L108

and D7 although i dont think so have this prob on d7?
https://github.com/pantheon-systems/drops-7/blob/853d658a900b376d6eaf81718b60573743c0beec/includes/bootstrap.inc#L764

PR #2547, opened 2020-08-04
Issue #2504: Skip permissions hardening by default on Pantheon apps

dustinleblanc:
I had some weird issues with running Drush on Drupal 8 with the default Pantheon settings.php files, but I managed to get it working and can confirm this works:
image


pirog:
@dustinleblanc this looks good but do we want to add a test to our Drupal 8 pantheon example? Seems like that would be easy.

pirog:
@dustinleblanc also should update the changelog i think!

pirog:
@dustinleblanc might be worth doing a `yarn generate:tests` locally to work through the escaping problem showing up in the tests

dustinleblanc:
@pirog definitely, I'll jump back on it after I finish the other auth stuff I am on

PR #2552, opened 2020-08-06
Fix bug causing lando pull to inadvertently pull cache data for panth…

pirog:
@dustinleblanc writing some tests for this real quick. 

I think adding the quotes here was my suggestion so MY BAD i guess ;)

### DIFF
--- a/docs/help/2020-changelog.md
+++ b/docs/help/2020-changelog.md
@@ -6,6 +6,7 @@ Lando is **free** and **open source** software that relies on contributions from
 
 * Fixed bug causing `db-import` wipe to fail on `views` [#2516](https://github.com/lando/lando/pull/2516)
 * Fixed bug causing `db-import` wipe to fail on table names with hypens [#2478](https://github.com/lando/lando/pull/2478)
+* Fixed bug causing `lando pull` to not correctly exclude `cache` data for `pantheon` sites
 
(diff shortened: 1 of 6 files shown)


pirog:
So TWO THINGS about this PR that are super weird. This is for our `landobot-drupal7` which uses Drush 8.4

1. Setting `--extra=--column-statistics=0`

```
lando terminus remote:drush -- sql-dump --structure-tables-list="cache_*"
```

Unless you added `--extra=--column-statistics=0`. Maybe pantheon sets that automatically now? We should double check on that. 

2. `--extra=--column-statistics=0` +  `--structure-tables-list="cache_*"` = bad

```
# Dumps just the cache tables structure
lando terminus remote:drush -- sql-dump --structure-tables-list="cache_*"

# Dumps BOTH the structure and data
lando terminus remote:drush -- sql-dump --structure-tables-list="cache_*" --extra=--column-statistics=0
```

expected or bug?


@dustinleblanc i think we should get some feedback from Pantheon (Greg maybe?) to make sure before we merge.

pirog:
Removed the tests here https://github.com/lando/lando/pull/2552/commits/d8244fa84d482e71db247c4d70a0137dd047b0b3 because i forgot the `remote:drush/wp` doesnt work when run through leia :(

PR #2558, opened 2020-08-09
2403 terminus auth

pirog:
@dustinleblanc PRing this as a draft so we can collaborate to get it to the finish line. I'll drop relevant comments in the code explaining the changes that we needed to make. Basically we needed to combine a bunch of things we tried to do singularly plus a few other things. Outside of that i think we want to do the following things to get this to the finish line:

- [x] Swap out any other references to `TERMINUS_TOKEN` (i think there is one in utils.js)
- [x] Manually test out the new interactive flow since we dont have a good way to test this in Leia eg does
- [x] We should probably provide some kind of warning message if the users token goes invalid, the warning dialog stuff is in art.js if we want a "big" warning. The warning should inform user their token is now invalid, revoked, expired, etc and they need to provide a new token that has access to the site, might be good linking to the create a new token page and/or other relevant docs about authing

Looking forward the foundational weakness this issue laid bare is the fact that the pantheon recipe has grown iteratively since the Kalabox days and as such its both accumulated evolutionary baggage and if there was an initial/overarching structure/design to it originally it has since been lost to the sands of time.

For that reason we might want to accelerate our plans to rewrite the integration so it follows the pattern of the other integrations. I'll PR some scaffolding for that but i think we can basically put the new thing in `experimental/lando-pantheon` and use `--experimental` to toggle between the two until the new thing is ready to go.


integrations/lando-pantheon/app.js
```diff
@@ -36,9 +36,11 @@ module.exports = (app, lando) => {
...
-    // Add tokens and other meta to our app
-    app.pantheonTokenCache = 'pantheon.tokens';
-    app.pantheonTokens = lando.cache.get(app.pantheonTokenCache) || [];
-    app.terminusTokens = utils.getTerminusTokens(lando.config.home);
+    // Load in other stuff like tokens and other meta at the most opportune moment
+    app.events.on('pre-init', 1, () => {
+      app.pantheonTokenCache = 'pantheon.tokens';
+      app.pantheonTokens = lando.cache.get(app.pantheonTokenCache) || [];
+      app.terminusTokens = utils.getTerminusTokens(lando.config.home);
+    });
```
pirog, line 44: @dustinleblanc this was actually the big roadblock. The relevant `pull|push|switch` commands were using the stuff populated in these vars and these were all loading BEFORE the stuff we wanted to do. Wrapping them so they only load during an app.init event was the money move.

integrations/lando-pantheon/index.js
```diff
@@ -1,6 +1,44 @@
...
+              const filteredTokens = _(globalTokens)
+                .filter(token => token.token !== cachedToken)
+                .value();
+
+              // Set or things correctly for downstream
+              lando.cache.set('pantheon.tokens', filteredTokens, {persist: true});
+              lando.cache.remove(_.get(data, 'options._app.toolingCache'));
+              lando.cli.clearTaskCaches();
+              // This is the money option
+              delete data.options.auth;
```
pirog, line 37: @dustinleblanc this was the other big thing we needed to do. As long as this is set the interactive auth question will never run because it things its been passed in by the user or already set as teh default value.

dustinleblanc:
I started to try to work on some of the code climate gripes but it was getting too tangled, might have to revisit later. All outlined changes should be in here though

Issue #2101, opened 2020-03-26
sysctl update for Elasticsearch not persisted after reboot


gabrys:
https://github.com/lando/lando/blob/b15eb178cfc95930f4e49e255a80c028abe6ac5b/installer/linux/scripts/postinst#L21-L24

To my understanding this will change the setting, but a reboot will make it go back to the default (or whatever the user has set previously).

A better solution would be to re-set this sysctl before Elasticsearch start which doesn't ever require sudo as you can mount /proc/sys/vm/max_map_count as a volume to the elasticsearch container and then read and write to that file as root  inside the container. Additionally the code should first check if the value is below ES-required limit, or otherwise it may decrease the value.

pirog:
Great suggestion. Is that something you could PR?

smutlord:
@gabrys id like to try and implement this, can you provide more details into what you are thinking?

gabrys:
@smutlord, Two steps are required for this to work:

1. mount /proc/sys/vm/max_map_count as a volume, so that this file is visible in the ES container
2. update the start script of ES image to check the current value and update it if it's less than 262144

Not sure how this will work in MacOS or Windows where there might not be the /proc/sys/vm/max_map_count file on the host, so extra testing is needed here

Which part (1, 2, or extra check for MacOS/Windows) do you need some help with?

pirog:
@gabrys AFAIK this would only need to happen on Linux as the Docker VM on macOS/Windows has `max_map_count` set above `262144` by default

gabrys:
That's also my understanding, but still, I'm not sure what happens if you specify in compose config a mount that doesn't exist, etc.

pirog:
@gabrys i think what im saying is we would only define this mount and do the needed start script things if the user is on linux

rabauss:
I'm having this issue with a fresh Windows 10 2004 and a clean Docker 2.3.0.4 in wsl2 mode - solution was:
```
//open terminal or powershell
wsl -d docker-desktop
sysctl -w vm.max_map_count=262144
```

gabrys:
@rabauss my understanding is WSL2 is running Linux kernel so from Docker perspective the host system looks a lot like Linux. Probably the note above saying the default `max_map_count` value on Windows doesn't apply to WSL2. I don't have easy access to a Windows machine, but I would still try the "Linux" solution being:

* mount host's /proc/sys/vm/max_map_count to containers /tmp/max_map_count

[...]

PR #2559, opened 2020-08-09
Update elasticsearch.md

gitressa:
Added clarification that `vm.max_map_count` needs to be defined on the host machine. I ran into this, even using the official release of Lando  v3.0.10 on Ubuntu.


pirog:
@gitressa this is actually a bug, see: https://github.com/lando/lando/issues/2101. Current goal is to get lando to work in the way described there so no additional setup step or docs are required.

gitressa:
Thanks for clarifying @pirog, I'll close this PR, and let it be fixed over there :)

PR #2567, opened 2020-08-12
update cookie stripping

### LINKED ISSUE #2566: Pantheon VCL unsets cookies improperly
VCL unsets cookies for full request strings that end in file types like jpg or gif.
This has been resolved in the upstream.
Example:
`/asdf.php` -- cookies operate with same rules as pantheon
`/file.jpg` -- cookies stripped (as expected)
`/asdf.php?file=file.jpg` -- cookies stripped, not expected
`/asdf.php?file=file.jpg&1=1` -- cookies not stripped

nullvariable:
change cookie filtering to only operate against the path, and bring extensions inline with Pantheon's current config

fixes #2566

### DIFF
--- a/integrations/lando-pantheon/recipes/pantheon/pantheon.vcl
+++ b/integrations/lando-pantheon/recipes/pantheon/pantheon.vcl
@@ -79,7 +79,7 @@ sub vcl_recv {
   # is so that developers always have fresh content while editing files.
   #
   # This list of file extensions is also in vcl_backend_response(), keep them in sync.
-  if (req.url ~ "(i?)\/(?!\?)[\w\-\=\&]+\.(png|gif|jpeg|jpg|ico|bmp|tif|tiff|webp|swf|css|js|woff|woff2|svg|ttf|otf|eot)($|\?)") {
+  if (req.url ~ "(i?)\/(?!\?)[\w\-\&~\.]+\.(png|gif|jpeg|jpg|ico|bmp|tif|tiff|webp|swf|css|js|woff|woff2|svg|ttf|otf|eot)($|\?)") {
     unset req.http.Cookie;
   }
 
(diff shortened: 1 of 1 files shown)


nullvariable:
@dustinleblanc or @pirog this should work now. I started with a far to complicated regex, but ended up redoing and just stripping the query string before the regex. The regex matches what we do against the path on the platform as of the last update in Oct 2019.

Bonus, there are headers that show the cookies were removed. I thought about putting a docs link in another header, but figured that was overkill, any thoughts?

pirog:
@dustinleblanc i dont have a strong opinion re: @nullvariable's question. Might be worth rerunning the tests since they look busted? Doesn't seem like this change should break tests but 🤷

dustinleblanc:
Kicked off another test run. Re: adding a header with docs, I don't have strong feels either...which means to me, unless @nullvariable is really feeling like doing it, don't worry about it :)

nullvariable:
builds are failing, and it seems like a legit error, so I'm going to run tests locally and go ahead and give the rest of the VCL logic a once over with our current config

pirog:
@nullvariable im on the scene and can take this all the way to the promised land, looks like this was the edit we needed to make
```diff
diff --git a/integrations/lando-pantheon/recipes/pantheon/pantheon.vcl b/integrations/lando-pantheon/recipes/pantheon/pantheon.vcl
index df63822c..557ea415 100644
--- a/integrations/lando-pantheon/recipes/pantheon/pantheon.vcl
+++ b/integrations/lando-pantheon/recipes/pantheon/pantheon.vcl
@@ -253,7 +253,7 @@ sub vcl_backend_response {
   # Don't allow static files to set cookies. This list of file extensions is also
   # in vcl_recv(), keep them in sync.

-  set bereq.http.url_no_qs = regsub(req.url, "\?.*$", "");
+  set bereq.http.url_no_qs = regsub(bereq.url, "\?.*$", "");
   if (bereq.http.url_no_qs ~ "(?i)\.(png|gif|jpeg|jpg|ico|bmp|tif|tiff|webp|swf|css|js|woff|woff2|svg|ttf|otf|eot)$") {
     set beresp.http.x-pantheon-removed-set-cookie = beresp.http.set-cookie;
     unset beresp.http.set-cookie;
```

pirog:
@nullvariable this was merged into another branch

Issue #2439, opened 2020-07-02
Lando tooling should take the option delimiters (--) into account


codebymikey:
Lando version: v3.0.5 on Windows 10 Professional

.lando.yml:
```yml
name: my-first-drupal8-app
recipe: drupal8
config:
  webroot: .
```

Lando tooling should take into account where `--` [option delimiters](https://unix.stackexchange.com/a/570733) are passed in, and stop processing the options, passing everything directly to the relevant tooling command.

This makes it difficult to run commands like `lando drush status -vv` which also share a similar `-vvv` argument without it also affecting lando.

We should instead be able to run it like:
```bash
lando -- drush status -vv
```
such that all the typical lando arguments are ignored, and everything else is passed into the relevant tooling script for processing.

pirog:
@codebymikey i dont think this is something we plan on doing. afaik the only place where the options are shared by both lando and the tooling command is `-vvv` and that is actually by design.

i also am not sure this format would play with with the options parser we use under the hood or for tooling commands that themselves require use of `--`

codebymikey:
> i also am not sure this format would play with with the options parser we use under the hood or for tooling commands that themselves require use of --

My thinking for this would be to detect the sole use of `lando --` directly (when tooling is involved) and act accordingly rather than all instances of `--`, as I can imagine that would break existing tools.
Such that:
- `lando -- drush -vvv` => `drush -vvv` (lando arguments aren't parsed)
- `lando -- drush -- status -vvv` => `drush -- status -vvv` (lando arguments aren't parsed)
- `lando drush -- status -vvv` => `drush -- status -vvv` (keeps existing behaviour, where `-vvv` also affects `lando`)

This approach shouldn't break existing tooling commands, and I think makes the tooling more flexible when the internal tooling shares similar arguments to Lando that you don't want Lando using, the main ones being `--help` and `--verbose`, but more obscure tooling might conflict with other Lando arguments in the future, and I think it should be possible to avoid those scenarios if needed.

I haven't looked at the tooling code in much detail, but I understand if it's not high priority especially if the code is particular complex.

Is this something you're looking to accept if it was implemented as a PR?

pirog:
@codebymikey i think we would merge a PR for this as long as:

1. This would be a secondary syntax for tooling commands, eg this would not change the existing tooling syntax. It probably also makes sense to document this syntax as an explicit way to separate args from `lando` and a subcommand. im not sure what other purpose it would serve..
[...]

PR #2572, opened 2020-08-20
#2439: Add POSIX "--" option processing functionality to tooling

codebymikey:
Hi @pirog,

Here's a PR addressing the #2439 issue with regards to [POSIX option processing](https://unix.stackexchange.com/a/570733).

I added both functional and unit tests, I think the unit test may be removed as I don't think it really does much.

The functional test may be tested by running:

```bash
yarn generate-tests && yarn mocha --timeout 900000 test/tooling-example.func.js
```

### DIFF
--- a/docs/config/tooling.md
+++ b/docs/config/tooling.md
@@ -236,6 +236,34 @@ lando word
 lando word --word=fox
 ```
 
+## Option processing
+
+It is possible to disable Lando option processing for tooling by passing `--` immediately after `lando`. This is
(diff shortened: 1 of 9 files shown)


codebymikey:
Still a useful feature to have as a CLI tool.

pirog:
@codebymikey agreed! we _do_ intend to support this but are revamping how the project is organized and _guessing_ it will make more sense to complete that work before we merge this in.

PR #2591, opened 2020-09-09
Drupal 9 is no longer in development

webchick:
Drupal 9 was released on June 3, 2020: https://www.drupal.org/docs/understanding-drupal/drupal-9-release-date-and-what-it-means

Removing inaccurate statement from the docs. Also, I just tested and the recipe worked fine for me, so it's possible this entire warning could be deleted too, but was less sure about that. :)


pirog:
thanks for the update @webchick! 

IIRC there are a few things we still wanted to do for D9 re: drush handling so i think we will keep the warning for now.

PR #2592, opened 2020-09-09
Fix broken link to Drush install docs

webchick:
In several places in the docs, there is reference to https://docs.drush.org/en/master/install/, however this page is a 404.

The new location seems to be https://www.drush.org/install/, so updated all instances to that instead.


pirog:
@webchick double thanks!

Issue #2603, opened 2020-09-14
unable to install on ARM


pbattino:
Are the Linux packages only available for x86?
I tried to install on a Debian running on an ARM board and I got an error message (I think it was "operating system not supported").

Is there a reason that prevents using Lando on ARM architecture? Or is it just that the packages haven't been done?

Thanks

rabauss:
I guess the problem is also the docker images - which are not built for ARM!

pbattino:
Sure, but there are also a lot of docker images built for ARM. I guess this is up to the developer to specify the right images for the project, instead of the default Lando ones, or am I missing something? 

(I'm just learning Lando, I have no real experience with it)

dustinleblanc:
@pbattino we definitely haven't created any arm images, and this is common issue with Docker. You have to maintain separate images for ARM and switch them, which we just haven't done because like... 99.9 percent of users are using an X86 processor for development machines.

Apple Silicon is definitely going to change the marketplace of what developers are running on their local systems, but that wave is a bit farther off still. For now you'll have to stick with X86 based hardware.

pbattino:
> You have to maintain separate images for ARM and switch them

Not sure but I think you don't need to  maintain separate images and switch them any more: docker now supports multi-architecture images: https://www.docker.com/blog/multi-arch-images/
There's a bit more work involved in creating the image, but for the user it should be completely transparent, as Docker should download the correct version automatically.

Anyway at the very minimum **this should be documented**! 
There's no mention of only one architecture supported in the System Requirements: https://docs.lando.dev/basics/installation.html#hardware-requirements
As far as Docker is concerned, the only requirement listed is Docker Engine, but that runs perfectly fine in architectures other than x86. Reading that, I assumed Lando would work on ARM, I had no reason to think otherwise.

It's a pity! I tried Lando on my laptop and liked it, but I do use a series of ARM boards for development work, I guess I need to stick to my current workflow.

dustinleblanc:
Noted @pbattino, I just opened a PR to document the requirement. Over the years that the project has existed, I just haven't seen anyone try to run Lando on anything not x86 yet. Raspberry Pi's are just about the only device that I would have imagined someone trying it on, but they don't have enough power for people to consider them for most web dev situations. The world is changing however and arm powered stuff is going to be a part of the future, even in consumer/business productivity hardware.  Thanks for the note about the multi-arch images!

pbattino:
Thanks for the request to update the doc! 
[...]

PR #2604, opened 2020-09-14
Issue #2603: Note that only x86 will work with our Docker images

dustinleblanc:
This just calls out processor architecture explicitly in the installation docs. When Apple Silicon launches, we'll have to see how that plays with Docker/Lando too.


pirog:
@dustinleblanc merged into https://github.com/lando/lando/pull/2609

PR #2673, opened 2020-11-03
Tell lando to use composer installed by adding composer_version

TheSETJ:
Thank you so much for contributing code to Lando! If this is your **first time** contributing to Lando we recommend you first check our [Getting Involved](https://docs.lando.dev/contrib/contributing.html) docs.

It may also be useful to check the below documentation based on what you are trying to do:

* [Adding code to Lando](https://docs.lando.dev/contrib/contrib-intro.html)
* [Updating documentation](https://docs.lando.dev/contrib/contrib-intro.html)
* [Working on DevOps](https://docs.lando.dev/contrib/contrib-intro.html)
* [Working on Lando's auxiliary services eg websites, apis, etc](https://docs.lando.dev/contrib/contrib-intro.html)
* [Writing Lando Guides](https://docs.lando.dev/contrib/guides-intro.html)
* [Writing Lando blog posts](https://docs.lando.dev/contrib/blogging-intro.html)

- [ ] ~My code includes an update to the [CHANGELOG](https://github.com/lando/lando/tree/master/docs/help)~ (Not required IMO)
- [ ] ~My code includes a [functional test](https://docs.lando.dev/contrib/contrib-testing.html) if applicable~ (Not applicable)
- [ ] ~My code includes a [unit test](https://docs.lando.dev/contrib/contrib-testing.html) if applicable~ (Not applicable)

### DIFF
--- a/docs/config/backdrop.md
+++ b/docs/config/backdrop.md
@@ -83,6 +83,15 @@ config:
   composer_version: '1.10.1'
 ```
 
+And then update your tooling, so lando use your preferred version of composer:
+
+```yaml
+tooling:
+  composer:
+    service: appserver
+    cmd: '/usr/local/bin/php /usr/local/bin/composer'
+```
+
 ### Choosing a web server
 
 By default, this recipe will be served by the default version of our [apache](./apache.md) service but you can also switch this to use [`nginx`](./nginx.md). We *highly recommend* you check out both the [apache](./apache.md) and [nginx](./nginx.md) services before you change the default `via`.
--- a/docs/config/drupal6.md
+++ b/docs/config/drupal6.md
@@ -84,6 +84,15 @@ config:
   composer_version: '1.10.1'
 ```
 
+And then update your tooling, so lando use your preferred version of composer:
+
+```yaml
+tooling:
+  composer:
+    service: appserver
+    cmd: '/usr/local/bin/php /usr/local/bin/composer'
+```
+
 ### Choosing a web server
 
 By default, this recipe will be served by the default version of our [apache](./apache.md) service but you can also switch this to use [`nginx`](./nginx.md). We *highly recommend* you check out both the [apache](./apache.md) and [nginx](./nginx.md) services before you change the default `via`.
--- a/docs/config/drupal7.md
+++ b/docs/config/drupal7.md
@@ -84,6 +84,15 @@ config:
   composer_version: '1.10.1'
 ```
 
+And then update your tooling, so lando use your preferred version of composer:
+
+```yaml
+tooling:
+  composer:
+    service: appserver
+    cmd: '/usr/local/bin/php /usr/local/bin/composer'
+```
+
 ### Choosing a web server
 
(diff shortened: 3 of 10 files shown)


TheSETJ:
A little backstory: when I tried this config, I noticed lando is using composer v2. That cause I wonder why lando ignores `composer_version` in action though I could see outputs related to installing my preferred version. After putting some time into it I found out that I need to redefine composer in my tooling. That why I think `composer_version`-related documentation pages are incomplete without saying how to use the preferred version instead of the default version shipped with lando, so I added it to them.

@pirog

TheSETJ:
Any update?

@pirog

pirog:
@TheSETJ i dont think we want to make this change. Right now the choice of what `composer` lando uses (or any tool for that matter) is governed by `PATH` See: https://docs.lando.dev/config/php.html#path-considerations This is also how tools are generally "chosen" in a non-Lando context.

This means that Lando is going to use the version in the the Landofile by default. If its not using that version then its almost always because the user has installed another version, via build step or otherwise, somewhere higher up in the `PATH`. 

We think this is a good default that works basically across projects and application types. Your proposed change is what we generally recommend to users who want to avoid `PATH` as a thing but that is generally the exception not the rule.

TheSETJ:
> @TheSETJ i dont think we want to make this change. Right now the choice of what `composer` lando uses (or any tool for that matter) is governed by `PATH` See: https://docs.lando.dev/config/php.html#path-considerations This is also how tools are generally "chosen" in a non-Lando context.
> 
> This means that Lando is going to use the version in the the Landofile by default. If its not using that version then its almost always because the user has installed another version, via build step or otherwise, somewhere higher up in the `PATH`.
> 
> We think this is a good default that works basically across projects and application types. Your proposed change is what we generally recommend to users who want to avoid `PATH` as a thing but that is generally the exception not the rule.

I proposed this change to lessen developer's confusion when they want to use a custom version of the composer. I myself experienced this and I know another developer that had the same experience. If the change I proposed is not suitable, then please suggest a replacement, because I think the documentation does not provide enough information about using a custom composer version to prevent the developer from confusion and wasting their time.

pirog:
I think the point I am making is that the proposed change is actually MORE
confusing because it diverges from how how lando sets other kinds of
tooling and also how command invocation works generally.  It’s also worth
pointing out that this case IS documented in so far as the PATH is
documented and that tooling overrides are documented.

We cannot document every edge case from everywhere otherwise the docs are
going to be too bloated to be useful which is ultimately going to waste
wayyyy more developer time.

If there is a replacement it would be to define an entire section on
choosing the composer version, similar to the other sections on choosing
stuff and then linking to the PATH considerations and how to override the
tooling if they experience what you are experiencing. It should definitely
not recommend that all users modify their tooling in the way that is being
proposed.

TheSETJ:
@pirog Maybe I'm missing something or the way I'm presenting my purpose is not completely clear; I'm just saying IMO documentation needs to give a hint to the developer to save them from confusion. It can be as simple as this: "Take a look at X if you have such an issue". By the way, as I said, maybe I'm missing something, so I don't try to force what I think.
Thanks for answering. :)

PR #2706, opened 2020-11-23
Update README.md

blymemikkel:
Add of platform.sh as an integration-able hosting provider


pirog:
@blymemikkel closing this because i merged in the change via another PR

PR #2714, opened 2020-11-27
Use new setting instead of deprecated config for temp path

Berdir:

This just updates the default overrides to use the new, non-deprecated setting for Drupal 8.8+: https://www.drupal.org/node/3039255

The old key won't work in Drupal 9.

### DIFF
--- a/integrations/lando-platformsh/lib/overrides.js
+++ b/integrations/lando-platformsh/lib/overrides.js
@@ -3,7 +3,7 @@
 const overrides = {};
 
 // DRUPAL 8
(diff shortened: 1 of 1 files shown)


pirog:
@Berdir seems like we'd want to keep the older one as well for backwards compatibility or am i not understanding this change?

Berdir:
Yeah, that's an option too, shouldn't hurt. \Drupal\Core\File\FileSystem::getTempDirectory will only trigger a deprecation message if the settings key is not set, so setting both is fine. Although any supported core version does support this now.

pirog:
@Berdir i'd make that change to the PR then and i can merge it in

PR #2719, opened 2020-12-01
2660/php8 updates

mikemilano:
Added docs & set default PHP version from 7.3 to 7.4


pirog:
@mikemilano merged this in over here https://github.com/lando/lando/pull/2721

Issue #2728, opened 2020-12-04
Xdebug 3 XDEBUG_CONFIG and XDEBUG_MODE environment variables prevents changing Xdebug config without rebuild


tbfisher:
f6bf4944724ae45eeed28e2900a021b8b7aec015 added `XDEBUG_CONFIG` and `XDEBUG_MODE` env vars.

I want to be able to change my php ini on the fly via tooling and change `xdebug.mode` to another value, but the `XDEBUG_MODE` env variable overrides this. Same for other values in the `XDEBUG_CONFIG` env var.

To reproduce

1. Set `xdebug: "debug,develop"` on your php-fpm service.
2. Start up and verify step debug.
3. Change the value via an ini file to `off` or to `develop` for example, restart the php-fpm service, verify the change.
4. The step debug still works.

dustinleblanc:
@tbfisher could you accomplish the same thing by using `export` to override the env var?

tbfisher:
@dustinleblanc for CLI yes, but not for HTTP afaik.

tbfisher:
I guess I should have specified that the env variables are set such that they override ini settings for both CLI and HTTP requests.

jcandan:
Could we swap the order of the merge items below to allow these environment variables to be overridden?

```diff
-        environment: _.merge({}, options.environment, {
+        environment: _.merge({}, {
           PATH: options.path.join(':'),
           LANDO_WEBROOT: `/app/${options.webroot}`,
           XDEBUG_CONFIG: `remote_enable=true remote_host=${options._app.env.LANDO_HOST_IP}`,
           XDEBUG_CONFIG: xdebugConfig(options._app.env.LANDO_HOST_IP),
           XDEBUG_MODE: (options.xdebug === true) ? 'debug' : options.xdebug,
-        }),
+        }, options.environment),
```

jcandan:
Tested locally. Happy to report that this change allows us to do the following again:

```yml
config:
  xdebug: false
services:
  appserver:
    overrides:
      environment:
        XDEBUG_MODE: debug
tooling:
  xdebug-on:
    user: root
    service: appserver
    description: Enable Xdebug.
    cmd: "docker-php-ext-enable xdebug && /etc/init.d/apache2 reload"
  xdebug-off:
    user: root
    service: appserver
    description: |-
      Disable Xdebug.
      (Recommended for performance when not actively debugging)
[...]

PR #2730, opened 2020-12-04
swap xdebug env var merge

jcandan:
Allow environment variable overrides to xdebug config.

### DIFF
--- a/plugins/lando-services/services/php/builder.js
+++ b/plugins/lando-services/services/php/builder.js
@@ -150,12 +150,12 @@ module.exports = {
       // Build the php
       const php = {
         image: `devwithlando/php:${options.version}-${options.image}-2`,
-        environment: _.merge({}, options.environment, {
+        environment: _.merge({}, {
           PATH: options.path.join(':'),
           LANDO_WEBROOT: `/app/${options.webroot}`,
           XDEBUG_CONFIG: xdebugConfig(options._app.env.LANDO_HOST_IP),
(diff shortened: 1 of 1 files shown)


jcandan:
closes #2728

pirog:
@jcandan AFAIK this change is not necessary but maybe i dont get what you are trying to do. See https://github.com/lando/lando/issues/2728#issuecomment-739988384 for more info.

codebymikey:
@pirog, I think what he was trying to do was allow users to override environment configurations like that without needing to do a full rebuild which can sometimes be expensive/time-consuming.

However I don't think this PR in particular achieves that.
I haven't had a chance to delve into the code yet, but do you know if this is something that's even feasible? - i.e. changing environment variables and getting them reflected without rebuilding, just like `docker-compose` does.

pirog:
We use the same mechanism that Docker compose does so the rebuild
requirement comes from there

pirog:
@codebymikey in order for environment variables to be `reset` docker needs to "recreate" the containers. This means destroy and then create. Lando only does this on a `lando rebuild` because we do not want to "recreate" containers unless the user has explicitly asked us to do so. However, this also is going to trigger lando to run its build steps, which are _not_ the same as `docker compose up --build` so if you have complex or time intensive build steps you are going to pay a big price for a simple envvar change. 

Eventually we are going to move into a situation where build steps can be cached so they do not need to rerun on a `lando rebuild` unless they have been changed. This will drastically speed up rebuilds and make changing envvars less of a big deal.

PR #2741, opened 2020-12-10
Update lando-with-vscode.md

RichardDavies:
Updated config with Xdebug 3 changes and added advanced setup section with custom tooling to toggle xdebug on and off.

### DIFF
--- a/docs/guides/lando-with-vscode.md
+++ b/docs/guides/lando-with-vscode.md
@@ -65,9 +65,10 @@ xdebug.max_nesting_level = 256
 xdebug.show_exception_trace = 0
 xdebug.collect_params = 0
 ; Extra custom Xdebug setting for debug to work in VSCode.
-xdebug.remote_enable = 1
-xdebug.remote_autostart = 1
-xdebug.remote_host = ${LANDO_HOST_IP}
+xdebug.mode = debug
+xdebug.client_host = ${LANDO_HOST_IP}
+xdebug.remote_port = 9003
+xdebug.start_with_request = trigger
 ; xdebug.remote_connect_back = 1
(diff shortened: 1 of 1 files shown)


pirog:
@RichardDavies this is good but i would recommend making it explicit that `9003` is for XDEBUG 3 and then note that XDEBUG 2 and below will still use port `9000`. What do you think?

RichardDavies:
> 
> 
> @RichardDavies this is good but i would recommend making it explicit that `9003` is for XDEBUG 3 and then note that XDEBUG 2 and below will still use port `9000`. What do you think?

Anyone following these steps has to do a `lando rebuild` as part of the process, so they will automatically pick up xdebug 3 from the updated Docker images now. So including legacy details doesn't seem necessary to me at this point.

pirog:
Agreed in principle. Including legacy details, however, probably will save us a bunch of time in support requests since there are still thousands of people on php 7.1 and lower who are still going to be stuck on xdebug 2.

RichardDavies:
So if you pin your PHP version in Lando to 7.1 or lower, then the docker image loads xdebug 2, even after running `lando rebuild`? I don't have a problem with keeping the legacy info if necessary, I'm just trying to understand how anyone would still be on xdebug 2 after running `lando rebuild` if the new Docker images contain xdebug 3 as you told me in Slack.

pirog:
That's right. 

XDEBUG 3 requires php 7.2 or higher (see the deps at the bottom of https://pecl.php.net/package/xdebug) so all images below that php version are going to remain on xdebug 2 presumably forever.

RichardDavies:
Done. The documentation has been updated with appropriate config and callouts for legacy xdebug support.

Issue #2762, opened 2020-12-19
Lando Pull and Lagoon Drupal File Directories


owenbush:
**Tell us about your setup**

Lando v3.0.23 on MacOSX Catalina

**Tell us about your `.lando.yml`**

```yml
name: dcoag
recipe: lagoon
config:
  build:
    - composer install
```

`.lagoon.yml`

```yml
docker-compose-yaml: docker-compose.yml

project: dcoag

tasks:
  post-rollout:
    - run:
        name: IF no Drupal installed and we are not master branch, sync from master branch
        command: |
            if tables=$(drush sqlq 'show tables;') && [ -z "$tables" ] && [[ "$LAGOON_GIT_BRANCH" != "master" ]]; then
                # no drupal installed, sync DB from master branch
                drush -y sql-sync @master @self
            fi
        service: cli
        shell: bash
    - run:
        name: drush cim
        command: drush -y cim
        service: cli
    - run:
        name: drush updb
        command: drush -y updb
        service: cli
    - run:
        name: drush cr
        command: drush -y cr
        service: cli

environments:
  master:
    cronjobs:
      - name: drush cron
        schedule: "*/15 * * * *"
        command: drush cron
        service: cli

```

**Tell us about the command you were running**

```bash
lando start
lando pull # or lando pull -d dcoag-master -f dcoag-master
```

**Tell us about the error you got**

When attempting to pull files from an existing lagoon instance we end up getting an error stating that the `web/sites/default/files` directory does not exist. This is not surprising to me as we use `docroot` as our Drupal webroot directory, not `web`. Also, the fact that the public file path is configurable and may not always be `sites/default/files` is somewhat problematic.

The issue arises from [this line within the lagoon-pull.sh file](https://github.com/lando/lando/blob/72fa3dd4641cd8b6b8b23aa59e8d58d14939a3d9/integrations/lando-lagoon/scripts/lagoon-pull.sh#L158)

`web/sites/default/files` is hardcoded to the `drush rsync` command.

The error we get is:

```bash
> rsync: change_dir "/app//web/sites/default" failed: No such file or directory (2)
> command terminated with exit code 23
> rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1659) [Receiver=3.1.3]
> rsync: [Receiver] write error: Broken pipe (32)

In RsyncCommands.php line 82:

  Could not rsync from dcoag-master@ssh.lagoon.amazeeio.cloud:web/sites/default/files to web/sites/default
```

[...]

PR #2763, opened 2020-12-19
2762 - Lagoon Pull - Fix hardcoded public file paths and minor bugs

owenbush:
- Fix problem with hardcoded public file paths in lagoon-pull.sh
- Fix problem with duplicated LANDO_LAGOON_PROJECT in files alias if no database import selected
- Clean up output to prevent flooding the console with irrelevant or duplicative drush output during pull process
- Add additional debug/notice messaging where appropriate


owenbush:
Resolves #2762

pirog:
Nice! Thanks @owenbush. Tried this out and worked well except for this check

```bash
drush "@${LANDO_FILES_ALIAS}" status | grep -q "File directory path";
```

My Drush lists the thing as "Files, Public" and not "File directory path" so its tripping. My take there is its probably best to avoid grepping on a label since presumably that can change and/or vary. I think for right now it's probably fine to just query drush for the `DRUPAL_FILES_PATH` and use that. The overwhelming majority of the time i think this will be set, and when its not im guessing we will still get a relevant error message.

Obviously we can improve that part in the future if im wrong ;) 

I'm going to merge this in and then make that edit myself.

PR #2770, opened 2020-12-22
Update XDebug port in launch.json to match XDebug 3.x default port.

gabe-connolly:
The default port for XDebug has changed from 9000 to 9003 in XDebug 3.x, so the docs should be updated to reflect this.


pirog:
@gabe-connolly im going to close this because i think @RichardDavies has a PR open that does this. Might be worth collaborating with him to get that PR merged. Let me know if im reading this wrong and it is not covered by below PR
https://github.com/lando/lando/pull/2741

PR #2809, opened 2021-01-14
Update instructions for installing PHP extensions

mlocati:
Since version 1.1.11, my `install-php-extensions` script is able to install oci8 (and Oracle Instant Client).

So, what about simplifying the instructions to install it?


pirog:
@mlocati definitely works for me! let me try it out

Issue #2540, opened 2020-08-03
XDebug / WSL 2: Make xDebug work out-of-the-box.


ayalon:
Hi

I wrote a blogpost about my WSL 2 setup under Windows 10. It is blazing fast and combined with mutagen.io sync it has a better performance than on OSX. ([See blogpost here](https://www.liip.ch/en/blog/setup-drupal-and-lando-with-wsl2-on-windows))

Almost everything works out-of-the-box except xDebug. I pinned down the issue, but I have no easy solution to it.

`lando/plugins/lando-core/index.js`
'''
LANDO_HOST_IP: (process.platform === 'linux') ? ip.address() : 'host.docker.internal',
'''

Because WSL 2 is a native linux kernel running under Windows, the above line detects 'linux' as the platform and therefore the ip.address() of  WSL is taken. This will not work, because as compared to OSX, PHPStorm or any xDebugger will run on the Docker Host system. So the host ip should be 'host.docker.internal'

This means, this condition should be extend to:
''' pseudeocode
if (platform === linux && platform !== wsl 2) ? ip.address() : 'host.docker.internal',
'''

If this is correct; XDEBUG_CONFIG will be automatically correct:
`plugins/lando-services/services/php/builder.js`
```
LANDO_WEBROOT: `/app/${options.webroot}`,
XDEBUG_CONFIG: `remote_enable=true remote_host=${options._app.env.LANDO_HOST_IP}`,
```

Unfortunatly I have no idea how to detect WSL 2 from javascript. Maybe someone has an idea how to fix that.

ianrogren:
@ayalon thanks for this!  Been banging my head for a while on why I kept getting timeouts with xdebug on Lando and WLS2.  Setting the environment override to LANDO_HOST_IP: "host.docker.internal" fixed my issue!

marfillaster:
Setting the env XDEBUG_CONFIG to `remote_enable=true remote_host=host.docker.internal` should work. If you're using VSCode Remote php-debug extension, you need to set `.configurations[].hostname = 0.0.0.0` in `launch.json`.

GuyPaddock:
Here's what worked for me with IntelliJ/PHP Storm running outside WSL being able to debug a PHP app using the Pantheon recipe in Lando:
1. Installed WSL 2 with Ubuntu.
2. Installed Docker Desktop, with WSL 2 support turned on.
[...]

PR #2814, opened 2021-01-19
First draft of WSL2 guide

dustinleblanc:
Thank you so much for contributing code to Lando! If this is your **first time** contributing to Lando we recommend you first check our [Getting Involved](https://docs.lando.dev/contrib/contributing.html) docs.

It may also be useful to check the below documentation based on what you are trying to do:

* [Adding code to Lando](https://docs.lando.dev/contrib/contrib-intro.html)
* [Updating documentation](https://docs.lando.dev/contrib/contrib-intro.html)
* [Working on DevOps](https://docs.lando.dev/contrib/contrib-intro.html)
* [Working on Lando's auxiliary services eg websites, apis, etc](https://docs.lando.dev/contrib/contrib-intro.html)
* [Writing Lando Guides](https://docs.lando.dev/contrib/guides-intro.html)
* [Writing Lando blog posts](https://docs.lando.dev/contrib/blogging-intro.html)

- [ ] My code includes the latest code from the `master` branch


docs/guides/setup-lando-on-windows-with-wsl-2.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Get WSL2 rolling
+
+The first thing we need to do is enable WSL2. To do so...
+
+1. Open the start menu and search for "Turn Windows features on or off".
+2. In the resulting window, check 'Windows Subsystem for Linux' and 'Virtual Machine Platform' and then click 'ok'. This starts the system installed for WSL and will ask us to restart after it downloads the stuff it needs. This really only gives you the WSL1 setup. To get WSL2...
+3. Download a kernel update to get WSL2 going. see the (Microsoft docs on installing the kernel update)[https://docs.microsoft.com/en-us/windows/wsl/install-win10#step-4---download-the-linux-kernel-update-package].
+4. After you install the kernel update, launch a powershell and run `wsl --set-default-version 2` to default to WSL 2.
+5. Head to the Microsoft store and install a Linux distro. I am going to choose [Ubuntu 18.04 LTS](https://www.microsoft.com/en-us/p/ubuntu-1804-lts/9n9tngvndl3q), since it is the most common distro for Lando users and our handy (Hyperdrive)[https://github.com/lando/hyperdrive] supports it.
```
reynoldsalec, line 54: Is this the correct Ubuntu install on the Microsoft Store @dustinleblanc?

docs/guides/setup-lando-on-windows-with-wsl-2.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Jump into Hyperspace
+
+After installing the latest LTS Ubuntu, using Hyperdrive makes the rest extremely easy. We simply launch a bash shell in the WSL instance, follow the user setup process, and then follow the Hyperdrive instructions in your terminal.
```
reynoldsalec, line 58: When you say "follow the Hyperdrive instructions", you're just talking about the CLI prompts correct @dustinleblanc?

docs/guides/setup-lando-on-windows-with-wsl-2.md
dustinleblanc: I think so, but it has been a bit since I did this

docs/guides/setup-lando-on-windows-with-wsl-2.md
dustinleblanc: Yep, though the readme would also be a good source of information

docs/guides/setup-lando-on-windows-with-wsl-2.md
rabauss: Obviously, it is also possible to install it manually: https://docs.microsoft.com/en-us/windows/wsl/install-manual
But I also did the installation via store a while ago

docs/guides/setup-lando-on-windows-with-wsl-2.md
```diff
@@ -29,41 +29,47 @@ feed:
...
 
-1. **never** run a lando app from within the default directory that WSL drops you into (`/mnt/c/Users/whateveruser`). This is the Windows filesystem mounted into your WSL instance, and it suffers from the filesystem performance issues that this setup avoids, therefore making going through the trouble of setting this up pointless.
-2. You have to start Docker every time you start up the WSL2 instance. I just run `sudo service docker start` but you can probably do some `.bashrc` magic to do this automatically. This is because the WSL2 distros have no init system, so docker doesn't auto start.
+1. **Never** run a Lando app from within the default directory that WSL drops you into (`/mnt/c/Users/whateveruser`). This is the Windows filesystem mounted into your WSL instance, and it suffers from the filesystem performance issues that this setup avoids, therefore making going through the trouble of setting this up pointless.
+2. You have to start Docker every time you start up the WSL2 instance. I just run `sudo service docker start` but you can probably do some `.bashrc` magic to do this automatically. This is because the WSL2 distros have no init system, so Docker doesn't auto start.
```
rabauss, line 63: For me docker is also automatically started, if I run `lando start` for the first time after reboot

reynoldsalec: I guess that makes sense, since Lando should know how to initialize Docker from within a standard Ubuntu environment @rabauss. So unless you have a contradictory experience @dustinleblanc, perhaps this warning is unnecessary?

dustinleblanc: It has never worked for me, but it's possible that this has been fixed. I may have just gotten accustomed to doing this from before it worked properly

reynoldsalec (approved):

mrtronje:
@dustinleblanc sorry for writing here, but could you elaborate why it's neccessary to run docker inside WSL2 instead of windows docker with wsl2 backend? what were problems with wsl2 windows backend?
we run into some problems with that configuration, especially we could not get xdebug to work with phpstorm. If i read ddev documention they don't suggest to do the same and I found others that have a different lando+wsl2 setup (https://www.liip.ch/en/blog/setup-drupal-and-lando-with-wsl2-on-windows) .

dustinleblanc:
The reason is that it quite simply provides better performance in my testing. My initial setup was using the WSL2 Docker backend and it didn't much improve performance in any metric over just running Lando the 'classic' way on Windows. Battery life, heat generation, speed, all were abysmal. The moment I switched to using a 100% inside WSL setup (as outlined in the documentation) my system began behaving more or less just like a Linux laptop. No excess fan usage, no great impact to battery life, no thermo-nuclear meltdown in my lap, and `drush cr` runs that didn't require watching an episode of Modern Family to pass the time while they ran.

I've had the feedback from folks a few times now that this part of the guide confused them, so it seems like we need to add more details, but in my personal experience, the difference between these methods was a super obvious night and day improvement in performance in every dimension that I care about.

mrtronje:
@dustinleblanc thank you for your answer! 

>no thermo-nuclear meltdown in my lap  🤣 

we should keep in mind that this suggested setup doesn't work with other implementation in the lando code base. The problem with docker inside wsl2 instead of docker desktop for windows with wsl2 backend is that host.docker.internal isn't set. And so LANDO_HOST_IP isn't set correct for docker inside of wsl2 (see https://github.com/lando/lando/blob/06b6e65a3a65ae98db23ce753e928353d6478577/plugins/lando-core/index.js#L101).
This part was written in respect of using docker desktop with wsl2 backend. If we suggest to use docker inside of wsl2 we break existing code in a way. This could be important if we are trying to use xdebug (https://github.com/lando/lando/issues/2540) that relies on LANDO_HOST_IP.
Maybe we should think about to update the documentation or better think about a fix for setting LANDO_HOST_IP so that docker inside wsl2 works 😄

iluuu1994:
@dustinleblanc

> The reason is that it quite simply provides better performance in my testing. My initial setup was using the WSL2 Docker backend and it didn't much improve performance in any metric over just running Lando the 'classic' way on Windows.

That's weird because the two should be equivalent for the most part. Have you repeated your tests lately? Maybe it was early on in the WSL2 or Docker for Windows with WSL2 backend and there was some kind of bug? We notice literally no difference between the two.

> Uninstall Pre-existing Docker + WSL2 Backend You will want to make sure that, if you have an existing Docker or WSL setup, you are not using the Docker + WSL2 backend. If you have this setup, you're going to have a bad time . It is very sad, very frustrating, so if you are here, do yourself a favor and uninstall the Docker Desktop WSL2 backend.

I think the docs should at least be updated to be less dramatic. From this statement it's pretty unclear whether Lando runs at all or if random issues will pop up. But it turns out it works perfectly fine. We were scared off by the message, installed Docker in WSL2 manually but then ran into the networking issues mentioned above.

dustinleblanc:
@iluuu1994 yes it has been a while since I tried that, once it went south I never revisited the experience because it was so bad. Sounds like we should. The drama was intended to reflect the sheer number of folks who have tried to configure this on their own, went down the route of Docker-Desktop, and wound up with a system that wasn't behaving in one way or another, which is what prompted us to create an 'official' guide (rather than just continuing to point people to Cal Evans' blog post). We were repeatedly seeing issues with people trying to adopt WSL2 and having something not working, and time and again, they weren't following the method I used to set my system up. Microsoft seems to be rapidly developing WSL2 though as I've seen the system change several times while configuring it on different machines, so I am open to the possibility that things have changed.

@mrtronje Networking hasn't been an issue for me, but I also don't use XDebug. I did try to test out an XDebug documentation change on this setup and couldn't get it work, however I have _NEVER_ been able to get XDebug to reliably work, so I didn't think it was related to this setup.

All things considered, we've gotten a surprising amount of feedback about this guide since it was released, and updates are prudent. 

I am not sure when I'll have a chance to sit down with my Win 10 system and spend some time testing this stuff out again, I've shuffled the family computers around the house again so I am not currently using that system for work (I've instead reclaimed my desktop for full-time Linux usage and I am working with my M1 mac otherwise).

iluuu1994:
@dustinleblanc

Thanks for your response!

> We were repeatedly seeing issues with people trying to adopt WSL2 and having something not working, and time and again, they weren't following the method I used to set my system up.

I wasn't aware that there issues other than performance. Well, I think revisiting this would be worth a shot. Maybe we could provide detailed instructions of what we're doing. :slightly_smiling_face:

PR #2829, opened 2021-01-28
Updated Redis documentation to match code

jnous:
In `plugins/lando-services/services/redis/builder.js` Redis version is defined as `['6', '5', '5.0', '4', '4.0', '2.8']`. This causes an error "redis version 6.0 is not supported" if a user configures their Redis service as `type: redis:6.0`. I suggest matching the docs to code.

### DIFF
--- a/docs/config/redis.md
+++ b/docs/config/redis.md
@@ -12,7 +12,7 @@ You can easily add it to your Lando app by adding an entry to the [services](./.
 
 ## Supported versions
 
-*   [6.0](https://hub.docker.com/_/redis)
+*   [6](https://hub.docker.com/_/redis)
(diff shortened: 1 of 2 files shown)


pirog:
@jnous this is good, however if `6.0.` is a valid docker tag for redis, which _im guessing_ it is then this might be a case where we need to change the _code_ to match the _docs_. What do you think?

jnous:
@pirog a fair point, that would make sense. (I admit I didn't think further as the change was so easy to propose...)

pirog:
@jnous no worries. that's why they pay me the big bucks ;) 

and by big bucks i mean i get paid mostly in emojis

PR #2833, opened 2021-01-30
Update wkhtmltopdf version to match current version on Pantheon

NickDickinsonWilde:
Had a chat with Pantheon support the other day partially due to differences in output and yeah Pantheon's current wkhtmltopdf version is different than what Lando has.
0.12.5 has a nasty bug that it doesn't output Table of Contents... but probably best to match the Pantheon version despite that - since it lead to a ton of confusion at least in my case.


pirog:
@NickDickinsonWilde agreed! thanks for the PR!

pirog:
@NickDickinsonWilde sadly this does not actually work. Did you try this code change out at all?

Issue #2760, opened 2020-12-17
Expose more variables in the NGINX template


Xilonz:
Lando exposes environment variables to the system, however, these variables are not available in the NGINX template renderer used here: https://github.com/lando/lando/blob/master/plugins/lando-services/services/nginx/launch.sh

We'd like to build an NGINX config file that automatically proxy passes to a domain specified in the lando (env_file) config so we can use one single NGINX vhost config file across multiple projects.

In our case we're working on -a lot of- WordPress websites where we usually don't want to pull the uploads folder, we proxy pass that to the original site.

I've tried to parse the variables in the build_as_root on the service.

Our current solutions are:
- replace the launch.sh using overrides - command.
- add a dummy domain, replace the dummy domain in the config and reboot nginx.

I've tried to find out what the `render-template` command is, but it's not available in the container.

akasake:
@Xilonz We also depend on `render-template` to replace env variables. It seems it's a bitnami thing ( https://github.com/bitnami/render-template ).

In https://github.com/lando/lando/commit/e9e6af49b373132a1b4171412c407dfb75602102 the default nginx image is changed from `1.14` to `1.17`. When I change it back to `1.14` everything seems to work again.

Although `render-template` is also on the `1.17` image, it seems `launch.sh` can't find it.

--

So if you need this, make sure to use nginx version `1.14` for now.

Xilonz:
@pirog, I've confirmed @akasake's solution works.

I've also tried to up the version to 1.18 but no avail.

I'm not too familiar with Docker, but this file: https://github.com/bitnami/bitnami-docker-nginx/blob/master/1.18/debian-10/Dockerfile, doesn't seem to be executed at all. The commands listed as RUN don't seem to be installed. 
Ex: `chmod -R +x  /opt/bitnami && /opt/bitnami/scripts/libcomponent.sh && component_unpack "render-template" "1.0.0-3" --checksum 8179ad1371c9a7d897fe3b1bf53bbe763f94edafef19acad2498dd48b3674efe`, I cant get this working.

Running the docker image (e.g. `docker run -v bitnami/nginx:1.17`) directly does contain the render-template function.

PR #2852, opened 2021-02-07
Add the render template component

Xilonz:
This adds the render template component back to NGINX so we can use env variables inside our NGINX templates.

I'm not sure if (postunpack.sh)[https://github.com/bitnami/bitnami-docker-nginx/blob/master/1.19/debian-10/rootfs/opt/bitnami/scripts/nginx/postunpack.sh] is needed here.

This code is an (almost) direct copy from the (Bitnami docker file)[https://github.com/bitnami/bitnami-docker-nginx/blob/master/1.19/debian-10/Dockerfile].

I've added a conditional check to maintain compatibility with the 1.14 images. 

Relevant issue: https://github.com/lando/lando/issues/2760

I'm aware that we'd need to keep the render-template component up-to-date but this wouldn't be too big of an issue, would it?

### DIFF
--- a/plugins/lando-services/services/nginx/launch.sh
+++ b/plugins/lando-services/services/nginx/launch.sh
@@ -34,10 +34,17 @@ if [ -f "$PARAMS_SOURCE" ]; then
   cp -f "$PARAMS_SOURCE" "$PARAMS"
 fi
 
+# Unpack components
+if [[ -f "/opt/bitnami/scripts/libcomponent.sh" ]]; then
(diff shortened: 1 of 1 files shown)


pirog:
@Xilonz seems reasonable for the short term. 

Longer term i think the only "real" solution here is to probably get off the bitnami images and maintain our own set of auxiliary tools and scripts. We don't do that now almost exclusively because we dont have the capacity to handle the extra maintenance burden of that.

pirog:
@Xilonz merged over here https://github.com/lando/lando/pull/2875

PR #2856, opened 2021-02-12
Support mariadb 10.5 and update others

ElijahLynn:
- Update supported MariaDB minor versions - From https://hub.docker.com/r/bitnami/mariadb/tags. 
- Add MariaDB 10.5 to supported versions.


plugins/lando-services/services/mariadb/builder.js
```diff
@@ -8,12 +8,13 @@ module.exports = {
   name: 'mariadb',
   config: {
     version: '10.3',
```
ElijahLynn, line 10: Should we bump the default too? 

10.5 was released on Dec 3, 2019. 

https://mariadb.com/kb/en/mariadb-1050-release-notes/

ElijahLynn:
Tugboat PR upgrade here > https://github.com/department-of-veterans-affairs/va.gov-cms/pull/4303. 
Lando PR upgrade here > https://github.com/department-of-veterans-affairs/va.gov-cms/pull/4304 (blocked on upstream PR)

ElijahLynn:
From Slack:

@Elijah Lynn the test not passing for your PR are because of the changes in the PR, which is to say the test failure is expected in this case
2:43
1) should use 10 3 22 as the default version is what Circle CI reports as the busted test, presumably because you updated the image

Elijah Lynn  2:45 PM
@pirog Okay, I'll update that. So tests should be in a passing state then, is what I am hearing and I should be able to get all the tests to pass.

pirog  2:47 PM
yes, tests should be in a passing state, however sometimes tests can false negative. in these situations we (eg the lando team) will rerun the tests until they pass and sometimes we even accept PRs if certain tests dont pass
2:48
we’ve got a plan on how to improve testing significantly but its a big undertaking and tied to other initiatives all of which are closer to beginning than done

Elijah Lynn  4:17 PM
Thanks for confirming that tests should be in a passing state. That PR doesn't change the default version so I'm not sure why is says that just yet. (edited) 
New

pirog  5:11 PM
It’s because you updated the pinned image for 10.3  and it likely has a different patch version

Elijah Lynn  5:54 PM
Ahhhhhhhh, I see, gotcha.

Issue #2855, opened 2021-02-12
pantheon_binding constant not defined


cellar-door:
The Lockr plugin/modules look for the pantheon_binding constant to be defined in order to identify the Pantheon environment and look for the connection certificate. Recently, not sure when, this constant stopped being defined in the environment.

Running:

Pantheon Recipe
Lando 3.0.25
Mac Mojave
Nothing fancy in lando.yml other than xdebug settings

reynoldsalec:
@cellar-door when I run `lando ssh -s appserver -c env` I see `pantheon_binding` defined in the `PRESSFLOW_SETTINGS`. Is this still a problem for you guys?

PR #2859, opened 2021-02-12
Updates the location the binding.pem is downloaded to for Lockr support

cellar-door:
Small update to point the download of the binding.pem to the correct folder where Lockr can find it. This is related to https://github.com/lando/lando/issues/2855 in that without the PANTHEON_BINDING not being defined Lockr ceases to work, however we are moving off using that constant and this will be the fix for that issue.


nullvariable:
this looks good to me

pirog:
@cellar-door @nullvariable im going to close this in favor of https://github.com/lando/lando/pull/2876 which _I THINK_ will get us where we want to go. If i'm wrong about that i'll owe you both a beer 🍺

PR #2870, opened 2021-02-19
Attempt to remove prestissimo on all composer versions

codebymikey:
Prestissimo is only removed on composer branch aliases.

It should support the alternative format as well.

See https://github.com/composer/composer/blob/4e0b8c1/src/Composer/Console/Application.php#L494-L507


pirog:
@codebymikey yes. good call.

PR #2878, opened 2021-02-25
Lando pull/push match interactive labels to acli UX.

serundeputy:


experimental/lando-acquia/lib/api.js
```diff
@@ -80,8 +80,9 @@ module.exports = class AcquiaApi {
       const envs = [];
       _.each(this.environments, env => {
         if (env.name != 'prod') {
+          const name = `${env.label}, ${env.name} (vcs: ${env.vcs.path})`;
           envs.push({
-            'name': env.name,
+            'name': name,
```
mikemilano: This can just be `name` since both key and value are the same. Also, please remove quotes from the property keys. I think I introduced it in this case, but it's not normal for this codebase.

mikemilano (approved):

Issue #2910, opened 2021-03-15
"#!/usr/bin/env node --no-warnings" in `lando.js` fails on Fedora


jazty:
Pulled the latest from 'master' branch (on commit `57ce9e35becc46a2f5c003b44273fdb18b1ba2a8`) and encountered this error:
```
$ lando start                                                                                                                             130 ↵
/usr/bin/env: ‘"node --no-warnings"’: No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines
```
It seems that Fedora (and probably other GNU/Linux distros) don't support passing flags to commands invoked by `env`.

https://unix.stackexchange.com/questions/63979/shebang-line-with-usr-bin-env-command-argument-fails-on-linux

Is `--no-warnings` necessary?

`git log -p bin/lando.js`:
```
commit 547252a4fc417627bb6e5afb4e8a4cb388cb6ecc
Author: Mike Pirog <mike@lando.dev>
Date:   Tue Mar 9 12:50:31 2021 -0500

    #2688: Force DOCKER_BUILDKIT=1 for now

diff --git a/bin/lando.js b/bin/lando.js
index f6980fbb..d2a99de1 100755
--- a/bin/lando.js
+++ b/bin/lando.js
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env node --no-warnings
```

jazty:
This appears to work on Fedora:
`/usr/bin/env -S node --no-warnings`

`man env':
```
 -S, --split-string=S
              process and split S into separate arguments; used to pass multiple arguments on shebang lines
```

Can anyone confirm that this works on Darwin?

jazty:
Dang. Ubuntu with coreutils 8.28 doesn't support the `-S` option for `env`.

This appears to work in both Fedora and Ubuntu, however:
`/usr/bin/env NODE_OPTIONS=--no-warnings node`

jazty:
#2911

pirog:
@jazty i think we are just going to remove this and it will be up to the user to handle warnings if they install from source

PR #2911, opened 2021-03-15
Setting `node --no-warnings` in an alternate way

jazty:
Setting `node --no-warnings` in an alternate way that will (hopefully) work with all `env` implementations.


jazty:
Issue: #2910

jazty:
Nope. Further testing showed this failing.

Issue #1847, opened 2019-10-31
Possible to change COLUMNS environment variable?


gitressa:
The command line column width is set to 256 columns:
```
$ lando ssh -s appserver -c env | grep -i col
COLUMNS=256
```
Is it possible to change this? I found the  [Environment](https://docs.lando.dev/config/env.html) page, and tried tweaking the value by creating a new `defaults.env` file, and adding this:
`COLUMNS=80`

... and this to my `.lando.yml` file:
```
env_file:
  - defaults.env
```
... and rebuilding Lando, but it doesn't seem to work.

Is it even possible to change this? I would be fine setting this globally, for all my Lando environments.

gitressa:
stalebot seems keen to close this issue ...

gitressa:
It says on https://github.com/probot/stale#is-closing-stale-issues-really-a-good-idea

> A comment is all it takes to keep the conversation alive.

Yet, now it's closed ...

generalredneck:
Also looking for this feature.

gitressa:
Yes, but stale bot closed the issue two weeks ago, so the maintainers might not see this issue any longer. 

> pirog removed the wontfix label on Dec 3, 2019

... gives me the impression that this could potentially be looked on. Creating a new identical issue doesn't seem optimal.

pirog:
AFAIK this is already possible with either an environment file, service level overrides or setting a user space lando global config that overrides the `appEnv` key. 

If those don't work this is something we could look into at some point but i'm guessing it would fall lower on the priority list.

gitressa:
Thanks for your input @pirog. I went back and experimented some more. I could have sworn that I already tried this, but perhaps I forgot to rebuild ... but this works:

Set per project in `.lando.yml`, as a service level override:

```
services:
  appserver:
    overrides:
      environment:
        COLUMNS: "140"
```

...or set with a user space lando global config in `~/.lando/config.yml`:

```
appEnv:
  COLUMNS: "140"
```

Looking in the output of `lando config` helped me debug.

- Ubuntu 18.04
- Lando 3.0.0-aft.2
- Drush 9.7.2

I understand it is a matter of taste, but would it make sense to reduce the default number of columns from 256 to for example 180 in https://github.com/lando/lando/blob/master/plugins/lando-core/index.js? Looking at my own command line with `COLUMNS=238` makes the font quite small ...

reynoldsalec:
[...]

PR #2978, opened 2021-04-26
Updates CLI COLUMNS

gitressa:
CLI column width is set to 256 characters, which can make some lines quite long and hard to read, as well as break in non-optimal ways. For more see #1847.

> The optimal line length for your body text is considered to be 50-60 characters per line, including spaces (“Typographie”, E. Ruder). Other sources suggest that up to 75 characters is acceptable.

https://baymard.com/blog/line-length-readability

Perhaps we can set COLUMNS to 120 characters in stead, which results in about 80 characters for the description, in the right hand side?

**Now, 256 character columns**
```
$ drush help config:import
Import config from a config directory.

Arguments:
  [label] A config directory label (i.e. a key in \$config_directories array in settings.php).

Options:
  --preview[=PREVIEW] Deprecated. Format for displaying proposed changes. Recognized values: list, diff. [default: list]
  --source=SOURCE     An arbitrary directory that holds the configuration files. An alternative to label argument
  --partial           Allows for partial config imports from the source directory. Only updates and new configs will be processed with this flag (missing configs will not be deleted). No config transformation happens.
  --diff              Show preview as a diff.
```

**120 characters**
```
$ drush help config:import
Import config from a config directory.

Arguments:
  [label] A config directory label (i.e. a key in \$config_directories array in settings.php).

Options:
  --preview[=PREVIEW] Deprecated. Format for displaying proposed changes. Recognized values: list, diff. [default:
                      list]
  --source=SOURCE     An arbitrary directory that holds the configuration files. An alternative to label argument
  --partial           Allows for partial config imports from the source directory. Only updates and new configs
                      will be processed with this flag (missing configs will not be deleted). No config
                      transformation happens.
```

### DIFF
--- a/plugins/lando-core/index.js
+++ b/plugins/lando-core/index.js
@@ -11,7 +11,7 @@ const path = require('path');
 const defaults = {
   config: {
     appEnv: {
-      COLUMNS: 256,
+      COLUMNS: 120,
       LANDO: 'ON',
       LANDO_WEBROOT_USER: 'www-data',
       LANDO_WEBROOT_GROUP: 'www-data',


codebymikey:
I think this would make sense to be configurable (with a default value of `120`, or whatever makes the most sense).

Configuration can either be pulled in from the `~/.lando/config.yml` or from an environment variable `LANDO_COLUMNS` or `COLUMNS`.

reynoldsalec:
@gitressa reading through the issue I think this is still a good idea (thanks for your persistence!). I was trying to make sure it doesn't interfere with and Lando ascii presentations, and I don't think it does.

Would you mind recreating this PR within https://github.com/lando/cli now that the file is over there?

gitressa:
Sure @reynoldsalec! I have created [Reduces CLI column width from 256 to 120 characters](https://github.com/lando/cli/pull/75).

reynoldsalec:
Awesome, I will close this PR in favor of that one, thanks @gitressa!

PR #3115, opened 2021-08-24
Update 'From source' section in installation.md

uberhacker:

### DIFF
--- a/docs/basics/installation.md
+++ b/docs/basics/installation.md
@@ -197,25 +197,26 @@ Once you've completed the above then do the following:
 
 ```bash
 # Clone the Lando source
-git clone https://github.com/lando/lando.git
+git clone https://github.com/lando/cli.git lando-cli
 
 # Install its dependencies
-cd lando && yarn
+cd lando-cli && yarn
 
 # ON WINDOWS:
-# Add /path/to/source/bin to PATH
+# Find the current path with command prompt: cd or powershell: pwd
+# Add C:\path\from\above\bin to PATH
 # See: https://www.computerhope.com/issues/ch000549.htm
-# Remember to relaunch your shell so the PATH changes take effect
+# Remember to relaunch your terminal so the PATH changes take effect
 lando.js
 # Or invoke directly
-node /path/to/source/bin/lando.js
+node "C:\path\from\above\bin\lando.js"
 
 # ON POSIX:
 # Set up a symlink
 # NOTE: we use lando.dev as a convention but you can name it whatever
 # This allows you to run stable lando and source lando side by side
 sudo mkdir -p /usr/local/bin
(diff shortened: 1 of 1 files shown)


hotwebmatter (approved):
Animated gif of a bearded guy doing something cool with a computer in the 1980s

Nice work, computer guy

Image of ASCII art spelling out LGTM, an acronym for Looks Good To Me
/approved

hotwebmatter:
Actually, maybe not approved.

It seems like the bespoke path for Lando dependencies might work with the new [Hyperdrive](https://github.com/lando/hyperdrive).

```
$ grep -rn /usr/share/lando
test/env.spec.js:59:      expect(dockerBinPath).to.equal('/usr/share/lando/bin');
test/env.spec.js:85:      filesystem({'/usr/share/lando/bin/docker-compose': 'CODEZ'});
test/env.spec.js:87:      expect(composeExecutable).to.equal('/usr/share/lando/bin/docker-compose');
lib/env.js:40:      return '/usr/share/lando/bin';
```

I don't have time to look into this at the moment, but it may be that the whole idea is for Lando to keep its own pinned version of `docker-compose` off the PATH, so that a newer version can be installed (by `brew`, `apt`, `yum`/`dnf` or whatever) and called interactively from the shell, without requiring Lando to use a newer version than it can handle.

If that's the case, then bravo -- I think it could be a huge win for Lando if it manages its own dependencies.

uberhacker:
Thank you for the review @hotwebmatter!  Can you elaborate on the motivation for the `/usr/share/lando/bin/docker-compose` location when you get a chance @pirog?  My initial thought was why not use a location already in PATH?

pirog:
Because we want to be able to ensure we use a version of docker-compose
that works with lando while not forcing the user to use our version of
docker-compose for all of their non-lando Docker things.

uberhacker:
I guess that makes sense.  Should there be pre-flight checks to flag for the unsupported docker-compose version?

hotwebmatter:
Makes sense to me @pirog.

Allowing Lando to manage its own dependencies will prevent a whole class of errors like this one:

```
 ■ Using an unsupported version of DOCKER COMPOSE
   You have version 1.29.1 but Lando wants something in the 1.23.0 - 1.29.0 range.
   If you have purposefully installed an unsupported version and know what you are doing
   you can probably ignore this warning. If not we recommend you use a supported version
   as this ensures we can provide the best support and stability.
   https://docs.docker.com/compose/install/#install-compose-on-linux-systems
```

With `/usr/share/lando/` I can install the very latest `docker-compose` on my `$PATH` and use that for all my non-Lando Docker things without worrying whether the version on my `$PATH` will break Lando.

:tada:

uberhacker:
Alright, I restored the `/usr/share/lando/bin` path.  This should be ready for your review when you get a chance @pirog.

PR #3202, opened 2021-11-04
clarified the description

billdennen:
Removed Drupal version numbers and WordPress


reynoldsalec:
@billdennen note that the Acquia docs will be moving over to https://github.com/lando/acquia/blob/main/docs/usage.md shortly

billdennen:
@reynoldsalec  thanks - that doc still mentions WordPress in the description. Will that be removed?

reynoldsalec:
@billdennen easiest thing to do would be to submit this PR against that other repo, otherwise I'm guessing it'll get lost in the shuffle.

The public docs won't regenerate in either case until a new tagged release of Lando comes out, so you'll only see the change in the repo.

PR #3345, opened 2022-03-06
Update README.md

gitressa:
Updates formatting to Markdown, and also the links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Lando is and always will be FREE and OPEN SOURCE. As such it relies on generous 
 
 ### Advanced Configuration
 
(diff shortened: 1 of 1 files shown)


reynoldsalec:
@gitressa I think the HTML formatting was intentional to get the "nice" styling on the GitHub display of the content. In my use pattern I'd consider that more important than the readability improvements of pure Markdown when viewing the raw README file, since I rarely look at raw READMEs, but I'm sure preferences range all over the map on that :)

gitressa:
Ah yes, the centering :) ... I guess it's a matter of personal taste, but I was never a big fan of it, it made it harder to scan for me.

But you're right, the looks take precedence over formatting. Though I am getting to be a fan of Markdown over HTML, since it's faster to write and readable.

Would this update work?

reynoldsalec:
This does a great job of hitting both goals @gitressa! Only thing that threw me off was the logo size, so I decided to keep that as an HTML element (to apply the width), but I think everything else looks fine in Markdown. Also good excuse to update the Lando logo :)

gitressa:
Cool, thanks @reynoldsalec -- and yes, SVG is an improvement, much crisper!

